PR #17, opened 2019-02-19
Check CODEOWNERS

nashif:


myrdyr:
I tried running this on the Zephyr repository, picking a fairly digestible commit range (1 file change), and it takes a _LOT_ of time to run. It seems like the parsing itself is very heavy, as it basically verifies the entire `CODEOWNERS` file in each run. On my computer, it took about 7 minutes to execute.

1. You should probably check more files than just the ones being added, as files can be renamed, moved or deleted, and that could affect `CODEOWNERS` as well.
2. If the intention is to run this only for file path changes, then I suggest to not run the `parse_codeowners()` if there are no new files being added/removed/moved.
3. Consider optimizing the parsing function. If you don't need to verify every path every time, you could make the logic opposite to what it is now. For each file, check if the path to that file matches some entry in `CODEOWNERS`.

nashif:
> I tried running this on the Zephyr repository, picking a fairly digestible commit range (1 file change), and it takes a _LOT_ of time to run. It seems like the parsing itself is very heavy, as it basically verifies the entire `CODEOWNERS` file in each run. On my computer, it took about 7 minutes to execute.

i9:zephyr(CODEOWNERS_fixes): time ../ci-tools/scripts/check_compliance.py -c v1.14.0-rc1.. -m Codeowners
Running Codeowners tests...
1 Errors found

real	0m0.596s
user	0m0.501s
sys	0m0.095s

Note that you need the fixed CODEOWNERS file I submitted as a PR to the main tree.


> 1. You should probably check more files than just the ones being added, as files can be renamed, moved or deleted, and that could affect `CODEOWNERS` as well.

first steps....
We can only improve from here...

myrdyr:
Great, I wasn't aware of the modified file. Runs much quicker then.

carlescufi:
I am OK with this but @myrdyr does this work out-of-tree? I assume it does since I see no issues with it, but just checking.

myrdyr:
@carlescufi Yes, it works as long as the repository has a CODEOWNERS file.

nashif:
> @carlescufi Yes, it works as long as the repository has a CODEOWNERS file. No dependency on ZEPHYR_BASE this time.

not that you can always exclude tests you are not interested in, and the fact it runs in this repo mean that it should work with any.

marc-hb:
I found a number of issues, thanks for reviewing a new version at PR #54

PR #18, opened 2019-02-22
scripts: print test failures. Other logging and help text enhancements

marc-hb:
Invoke init_logs() which was dormant and reads os.environ["LOG_LEVEL"]
Add new -v INFO/DEBUG/... command-line option.
Set default log level to WARN.
Log all errors at the ERROR level.

--commits defaults to the last commit.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -387,9 +387,13 @@ def init_logs():
 
     :return:
     """
+
+    # TODO: there may be a shorter version with something like:
+    # logging.basicConfig(level=getattr(logging, args.loglevel))
+
     global logger
     log_lev = os.environ.get('LOG_LEVEL', None)
-    level = logging.INFO
+    level = logging.WARN
     if log_lev == "DEBUG":
         level = logging.DEBUG
     elif log_lev == "ERROR":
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -618,17 +635,23 @@ def main():
...
 
-    errors = 0
+    failed_cases = []
 
     if args.github and 'GH_TOKEN' in os.environ:
         errors = report_to_github(args.repo, args.pull_request, args.sha, suite, docs)
     else:
         for case in suite:
             if case.result and case.result.type != 'skipped':
-                errors += 1
```
ulfalizer: `n_errors` might be a tiny bit clearer.

scripts/check_compliance.py
```diff
@@ -568,8 +578,14 @@ def main():
 
     :return:
     """
+
+    init_logs()
     args = parse_args()
 
+    if args.loglevel:
+        logger.setLevel(level=args.loglevel)
+    # else: os.environ.get('LOG_LEVEL'); see init_logs()
```
ulfalizer: Maybe some of this stuff could be moved into `into_logs()`, since it has to do with initializing logging. Could have it take `args` as an argument in that case maybe.

marc-hb: Submitted, see new PR #19

PR #19, opened 2019-02-22
check_compliance.py: simplify init_logs()

marc-hb:
As suggested by Ulf Magnusson in the previous PR #18, move all the
logging initialization in the init_logs() function.

Remove the switch/case parsing the log level name because Python does
that for free.


ulfalizer:
@nashif @carlescufi 
Alright if I merge approved stuff into this repo by the way?

PR #21, opened 2019-02-25
checks: kconfig: Check for references to undefined Kconfig symbols

ulfalizer:
Use 'git grep' to look for the pattern CONFIG_[A-Z0-9_]+ at word
boundaries in all files. For each match, check that the corresponding
Kconfig symbol is defined.

This check is broad enough that it will always need some whitelisting,
e.g. due to token pasting (CONFIG_FOO_#x) and non-existent symbols used
as examples. The message printed on failures points out where the
whitelist is.

There's a few stragglers in the whitelist for small issues that are
about to be fixed (the issue IDs are included).

Skip the samples/ and tests/ directories for now. They contain Kconfig
files that aren't part of the main Kconfig tree, which triggers false
positives. Maybe something fancier could be done later to include those.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import collections
 import sys
 import subprocess
 import re
@@ -222,6 +223,7 @@ class KconfigCheck(ComplianceTest):
 
         self.check_no_undef_within_kconfig(kconf)
         self.check_top_menu_not_too_long(kconf)
+        self.check_no_undef_outside_kconfig(kconf)
 
     def parse_kconfig(self):
         """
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -211,6 +215,124 @@ def run(self):
...
+        if undef_to_locs:
+            undef_desc = "\n".join(
+                "{:35} {}".format(sym_name, ", ".join(locs))
+                for sym_name, locs in sorted(undef_to_locs.items()))
+
+            self.case.result = Failure("references to undefined Kconfig symbols",
+                                       "failure")
+            self.case.result._elem.text = \
+                "Found references to undefined Kconfig symbols. If any of " \
+                "these are false positives, then add them to " \
```
marc-hb: Can you move the UNDEF_KCONFIG_WHITELIST data into a separate, standalone file? From a change control perspective it should tied to the main repo and not embedded in this script. It would also make grepping for one of these symbols in the main repo very useful: instant indication that it is in the whitelist. It could also enable other uses of the whitelist like for instance easily extracting the issue number when any.

scripts/check_compliance.py
```diff
@@ -211,6 +215,124 @@ def run(self):
...
+
+            # [7:] removes the "CONFIG_" prefix
+            if sym_name[7:] not in defined_syms and \
+               sym_name not in UNDEF_KCONFIG_WHITELIST:
+
+               undef_to_locs[sym_name].append("{}:{}".format(path, lineno))
+
+        if undef_to_locs:
+            undef_desc = "\n".join(
+                "{:35} {}".format(sym_name, ", ".join(locs))
```
marc-hb: I don't understand what undef_desc is and where this particular sym_name comes from.

scripts/check_compliance.py
```diff
@@ -211,6 +215,124 @@ def run(self):
...
+        for line in grep_stdout.decode("utf-8").splitlines():
+            path, lineno, sym_name = line.split("\0")
+
+            # [7:] removes the "CONFIG_" prefix
+            if sym_name[7:] not in defined_syms and \
+               sym_name not in UNDEF_KCONFIG_WHITELIST:
+
+               undef_to_locs[sym_name].append("{}:{}".format(path, lineno))
+
+        if undef_to_locs:
```
```python
if not undef_to_locs:
    return

File: scripts/check_compliance.py [hunk 4]
```diff
@@ -211,6 +215,124 @@ def run(self):
...
+                                        cwd=self.zephyr_base)
+
+        grep_stdout, grep_stderr = grep_process.communicate()
+        if grep_process.returncode:
+            self.case.result = Error("error running 'git grep'", "error")
+            # grep_stderr will be 'bytes' here, which prints as b'oops'
+            self.case.result._elem.text = \
+                "'{}' failed with exit code {} and wrote {} to stderr " \
+                "(while searching for Kconfig symbol references)" \
+                .format(grep_cmd, grep_process.returncode, grep_stderr)
```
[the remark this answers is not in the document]
ulfalizer: Seems fine to me. One issue is that `stderr` was added to `CalledProcessError` in Python 3.5. I think we still support Python 3.4, at least on paper. @nashif?

Looking at [the documentation](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError), it seems `check_output()` might not capture `stderr` either, but only `run()` (which was added in Python 3.5).

scripts/check_compliance.py
```diff
@@ -211,6 +215,124 @@ def run(self):
...
+                "{:35} {}".format(sym_name, ", ".join(locs))
+                for sym_name, locs in sorted(undef_to_locs.items()))
+
+            self.case.result = Failure("references to undefined Kconfig symbols",
+                                       "failure")
+            self.case.result._elem.text = \
+                "Found references to undefined Kconfig symbols. If any of " \
+                "these are false positives, then add them to " \
+                "UNDEF_KCONFIG_WHITELIST in {} in the ci-tools repo.\n\n{}" \
+                .format(__file__, undef_desc)
```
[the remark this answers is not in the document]
ulfalizer: Each independent section has a

```
#
# Blah blah
#
```

I was thinking of splitting those out into separate functions, but didn't get around to it yet. You'd want to reuse at least the `kconf` object to avoid having to reparse stuff, but could pass it as a parameter.

scripts/check_compliance.py
ulfalizer: My plan was to have someone complain, and then suggest moving `ci-tools` back into the main repo. ;)

That's simpler than adding code to parse the file.

scripts/check_compliance.py
ulfalizer: It's a description of the undefined variables.

The high-level format is this:

```python
undef_desc = "\n".join(... for sym_name, locs in ...)
```

Can try this if you haven't run across generators before:

```python
" foo ".join(2*s for s in ("bar", "baz", "qaz"))
```

scripts/check_compliance.py
ulfalizer: Fixed

scripts/check_compliance.py
```diff
@@ -173,8 +174,9 @@ def run(self):
             self.case.result._elem.text = str(e)
             return
 
+
         #
-        # Look for undefined symbols
+        # Look for references undefined symbols within Kconfig files
```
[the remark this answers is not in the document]
ulfalizer: Fixed

scripts/check_compliance.py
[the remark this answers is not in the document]
ulfalizer: Hmm... also just remembered that the stuff in https://github.com/zephyrproject-rtos/ci-tools/pull/26 will make it less awkward, so I'll wait for that.

scripts/check_compliance.py
ulfalizer: Added a comment above `undef_desc` in the code.

ulfalizer:
Anticipating a comment about token pasting, you could get rid of some stuff from the whitelist by skipping references ending in `_`, or references on lines with `##` in them.

Not sure if it's common enough to be worth it though. None of those got added during the PR and the standalone script that preceded it.

scripts/check_compliance.py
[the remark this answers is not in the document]
ulfalizer: Hmm... I guess the `stderr=subprocess.STDOUT` might make it work still. Feels a bit gnarly to include `stderr` in the parsed output though, even if it'd probably always be empty in practice.

scripts/check_compliance.py
marc-hb: Well, moving to the main repo wouldn't get _all_ the points I listed.

"Parsing the file" would really just be reading it. This patch already adds other, more complex "parsing" instances already.

scripts/check_compliance.py
marc-hb: Got it, the `for` keyword at the start of the line (and the late hour...) confused me. How about moving the `for` at the end of the previous line instead? Neither pylint nor pep8 complains.

marc-hb: > Suggestions for a better name would be appreciated.

Something like `undef_failure_msg`?

scripts/check_compliance.py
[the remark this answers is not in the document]
marc-hb: > > This is dropping grep_stdout in case of error.

> Feels a bit gnarly to include stderr in the parsed output though

You just helped me realize the current code also drops grep_stderr in case of success, think warnings. Maybe merging stderr with stdout and breaking the stdout parser on purpose is actually a good thing? It would mean "warnings" aren't silently discarded anymore.

Another, .communicate() alternative is to treat warnings as errors with:

```python
if grep_process.returncode or grep_stderr:
   Error(...
```

PS: this is exactly why people still use shell scripts for simple tasks (_not_ this one)
For convenience: https://docs.python.org/3.4/library/subprocess.html

ulfalizer:
Separate PR for splitting things up: https://github.com/zephyrproject-rtos/ci-tools/pull/31

Will rebase this after that one gets in.

scripts/check_compliance.py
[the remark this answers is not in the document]
ulfalizer: I did the `or grep_stderr` thing.

scripts/check_compliance.py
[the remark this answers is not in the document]
ulfalizer: Rebased now, and put into a separate function.

scripts/check_compliance.py
ulfalizer: I'd also want to support comments and stuff.

Might not be needed at least, if `ci-tools` can be moved back. If it can't be, I could add the separate file then, as a workaround.

scripts/check_compliance.py
ulfalizer: It isn't the entire message, just the description of the undefined variables, so might be a bit misleading (unless the code is changed to build up the entire message in that variable, but think that'd be overkilling it).

I added a comment to clarify it though.

ulfalizer:
https://github.com/zephyrproject-rtos/zephyr/pull/14575 needs to go in first. It's the only undefined symbol reference currently.

scripts/check_compliance.py
```diff
@@ -269,12 +269,12 @@ def check_no_undef_within_kconfig(self, kconf):
         Checks that there are no references to undefined Kconfig symbols within
         the Kconfig files
         """
-        undef_ref_warnings = [warning for warning in kconf.warnings
-                              if "undefined symbol" in warning]
+        undef_ref_warnings = "\n\n\n".join(warning for warning in kconf.warnings
+                                           if "undefined symbol" in warning)
```
marc-hb: I miss the benefit(s) of this change

scripts/check_compliance.py
```diff
@@ -301,6 +289,18 @@ def check_top_menu_not_too_long(self, kconf):
...
+        Checks that there are no references to undefined Kconfig symbols within
+        the Kconfig files
+        """
+        undef_ref_warnings = "\n\n\n".join(warning for warning in kconf.warnings
+                                           if "undefined symbol" in warning)
+
+        if undef_ref_warnings:
+            self.add_failure("Undefined Kconfig symbols:\n\n"
+                             + undef_ref_warnings)
+
```
marc-hb: Nit: the commit message should be more explicit that the code hasn't changed at all, would save time.

scripts/check_compliance.py
ulfalizer: Gets rid of a temporary list, though I might be the only one who cares. :P

marc-hb: A list is a "higher level" / more structured type than the same list flattened to a string. That's the part that annoys me a bit but I might be the only one who cares :-)

ulfalizer: Yeah, I get what you're saying. Could still read it as "if there's any warning text" to have it make sense at least. :)

PR #22, opened 2019-02-28
check_compliance.py: fallback on result.message when no ._elem.text

marc-hb:
There's not always an _elem.text. To test:

 ZEPHYR_BASE=/whatever ./scripts/check_compliance.py -m Kconfig

- Before:

  AttributeError: 'NoneType' object has no attribute 'strip'

- After:

  Running Kconfig tests...
  1 Errors found
  ERROR   : Test Kconfig failed: Can't find Kconfig

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -737,7 +737,8 @@ def main():
         for case in failed_cases:
             # not clear why junitxml doesn't clearly expose the most
             # important part of its underlying etree.Element
-            errmsg = case.result._elem.text.strip()
+            errmsg = case.result._elem.text
+            errmsg = errmsg.strip() if errmsg else case.result.message
             logging.error("Test %s failed: %s", case.name, errmsg)
 
     sys.exit(errors)


ulfalizer:
I wonder if setting `self.case.result._elem.text` could be mandatory instead, e.g. by having two functions `ComplianceTest.error()` and `ComplianceTest.fail()` that take the title and the error message, and using those throughout the test suite. Would cut down on some copy-paste too.

Could do this for the existing case. Gives a better error message.

```diff
diff --git a/scripts/check_compliance.py b/scripts/check_compliance.py
index f77b94e..5145010 100755
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -143,6 +143,7 @@ class KconfigCheck(ComplianceTest):
         kconfig_path = os.path.join(self.zephyr_base, "scripts", "kconfig")
         if not os.path.exists(kconfig_path):
             self.case.result = Error("Can't find Kconfig", "error")
+            self.case.result._elem.text = kconfig_path + " not found"
             return
 
         sys.path.insert(0, kconfig_path)
```

marc-hb:
> Could do this for the existing case. Gives a better error message

I found a non Kconfig specific way to get the better error message.

PR #23, opened 2019-03-01
checks: codeowners: Fail on missing trailing / and missing paths

ulfalizer:
Fail the CODEOWNERS check if there are missing trailing slashes or
missing paths, instead of just printing a warning to stderr. Otherwise,
bad entries will probably keep getting added to CODEOWNERS.

Clarify the error message when / is missing too.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -280,11 +280,13 @@ class Codeowners(ComplianceTest):
                         if abs_path.endswith("/"):
                             abs_path = abs_path + "**"
                         elif os.path.isdir(abs_path):
-                            error("Wrong syntax: {}".format(abs_path))
+                            self.add_failure("Expected / after directory '{}' "
+                                             "in CODEOWNERS".format(path))
                             continue
                         g = glob.glob(abs_path, recursive=True)
                         if not g:
-                            error("Path does not exist: {}".format(path))
+                            self.add_failure("Path '{}' not found, in "
+                                             "CODEOWNERS".format(path))
                         else:
                             files = []
                             if not add_base:


marc-hb:
That's not my understanding of how this script works. Based on other checks  I think it should create some JUnit `self.case.result = Failure("...")` instead and not die before running other tests.

ulfalizer:
> That's not my understanding of how this script works. Based on other checks I think it should create some JUnit `self.case.result = Failure("...")` instead and not die before running other tests.

Yeah, bit brusque. I changed it to fail the test instead.

Still not great in that you can't report multiple failures from a test without doing some ad-hoc stuff to build the message. Had to work around that in the Kconfig test.

scripts/check_compliance.py
```diff
@@ -236,11 +236,15 @@ def parse_codeowners(self, git_root, codeowners):
...
+                            self.case.result = Failure(
+                                "Expected / after directory in CODEOWNERS: "
+                                + abs_path)
                             continue
+
                         g = glob.glob(abs_path, recursive=True)
                         if not g:
-                            error("Path does not exist: {}".format(path))
+                            self.case.result = Failure(
+                                "Path does not exist: " + path)
```
marc-hb: In other examples Failure() takes two arguments, why not here?

ulfalizer: Didn't spot it. Fixed.

ulfalizer:
> Building up a message seems OK to me, it's also consistent with for instance checkpatch (which does the aggregation itself but end result is the same)

The way I'd want to solve it is with a `self.fail(...)` method, which would handle building the string for multiple failures in the same test.

Maybe that should go in a separate PR though, since it's a wider refactoring.

scripts/check_compliance.py
```diff
@@ -236,11 +236,17 @@ def parse_codeowners(self, git_root, codeowners):
                         if abs_path.endswith("/"):
                             abs_path = abs_path + "**"
                         elif os.path.isdir(abs_path):
-                            error("Wrong syntax: {}".format(abs_path))
+                            self.case.result = Failure(
+                                "Expected / after directory in CODEOWNERS: "
+                                    + abs_path,
+                                "failure")
```
marc-hb: Nit: move to previous line? "pylint" is useful for this if you manage to ignore the unrelated noise.

ulfalizer: Think it's clearer that it's two arguments like this. [This guy](https://www.youtube.com/watch?v=wf-BqAjZb8M) is my spirit animal when it comes to "PEP 8-ifying" code and the like. :)

marc-hb:
> > Building up a message seems OK to me, it's also consistent with for instance checkpatch (which does the aggregation itself but end result is the same)

So I was in a good place to test this and I did. I just realized something I missed earlier sorry: this does remove all but the last error from the standard output.

> The way I'd want to solve it is with a self.fail(...) method, which would handle building the string for multiple failures in the same test. Maybe that should go in a separate PR though, since it's a wider refactoring.

Without entering refactoring territory, it should be easy to aggregate the error  messages in the for loop and create a single Failure() at the end of it.

ulfalizer:
> Without entering refactoring territory, it should be easy to aggregate the error messages in the for loop and create a single Failure() at the end of it.

It'd still get overwritten by the `self.case.result = Error("CODEOWNERS Issues", "failure")`. I'd rather solve it once and for all later, in a robust way.

ulfalizer:
The `self.fail(...)` thing will just work by the way. Tests won't have to care whether they generate one failure or many. It'll work either way.

ulfalizer:
I could always do the `self.fail()` thingy first and then rebase this on top of it though, if it's a blocker.

marc-hb:
I clearly didn't spend nearly as much time understanding the code as you did; at least not yet. What I just observe as a "naive user" is losing all errors on the terminal except one compared to the current code. This means I could fix that one error and push, only to be told by github that it's not actually working (since you do fix that part). So this needs at the very least a suggestion to keep re-running until no error are reported.

(Of course even better if you find a way to report all errors without rewriting the whole script)

ulfalizer:
@marc-hb @nashif 
PR for the helpers I had in mind: https://github.com/zephyrproject-rtos/ci-tools/pull/26

If those seem fine, I'll rebase this PR on top of it later. It removes the need to manually build up a string when there's multiple failures in a single test.

ulfalizer:
@marc-hb @nashif 
Updated to use the new status reporting helpers now.

marc-hb:
> The reporting is incompatible with other checks, the text visible to the user on github needs to be in the body of the failure and not the content of message attribute.

This is what I got when I just tested the current revision, does this look OK @nashif ?
```xml
	<testsuite errors="0" failures="1" name="Compliance" skipped="0" tests="1" time="0">
		<testcase classname="Guidelines" name="Codeowners">
			<failure message="Codeowners issues" type="failure">
Expected / after directory '/arch/arm' in CODEOWNERS
Expected / after directory '/arch/nios2' in CODEOWNERS</failure>
		</testcase>
	</testsuite>
```

ulfalizer:
> By the way the XML and console outputs look IMHO both better with this (unrelated) change:
> 
> ```diff
> diff --git a/scripts/check_compliance.py b/scripts/check_compliance.py
> index 53353f0894d9..9d704ab17cf9 100755
> --- a/scripts/check_compliance.py
> +++ b/scripts/check_compliance.py
> @@ -138,7 +138,7 @@ class ComplianceTest:
>              self.case.result._elem.text = ""
>  
>          # If there are multiple Failures, concatenate their messages
> -        self.case.result._elem.text += msg + "\n\n"
> +        self.case.result._elem.text +=  "\n" + msg
>  
>  
>  class EndTest(Exception):
> ```

Some of the tests have individual failures with messages that are several lines long (e.g. the test for references to undefined Kconfig symbols within Kconfig files). Would look too scrunched up for those.

PR #24, opened 2019-03-01
[DNM] Codeowners: warn the current files are always used

marc-hb:
For instance this warning explains why the fixed commit fails:

  ownersfix=c78be1bb7b79
  git checkout ${ownersfix}~1
  scripts/check_compliance.py -m Codeowners \
         -c ${ownersfix}~1..${ownersfix}

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -221,6 +221,9 @@ class Codeowners(ComplianceTest):
     def parse_codeowners(self, git_root, codeowners):
         all_files = []
         with open(codeowners, "r") as codeo:
+            if self.commit_range != CURRENT_COMMIT_RANGE:
+                logging.warning("Checking %s range against current %s file" %
+                                (self.commit_range, codeowners))
             for line in codeo.readlines():
                 if not line.startswith("#") and line != "\n":
                     match = re.match("([^\s]+)\s+(.*)", line)
@@ -614,6 +617,7 @@ def report_to_github(repo, pull_request, sha, suite, docs):
 
     return comment_count
 
+CURRENT_COMMIT_RANGE = "HEAD~1.."
 
 def parse_args():
     """
@@ -622,9 +626,9 @@ def parse_args():
     """
     parser = argparse.ArgumentParser(
         description="Check for coding style and documentation warnings.")
-    parser.add_argument('-c', '--commits', default="HEAD~1..",
-                        help='''Commit range in the form: a..[b], default is
-                        HEAD~1..HEAD''')
+    parser.add_argument('-c', '--commits', default=CURRENT_COMMIT_RANGE,
+                        help='Commit range in the form: a..[b], default is %s'
+                        % CURRENT_COMMIT_RANGE)
     parser.add_argument('-g', '--github', action="store_true",
                         help="Send results to github as a comment.")
 


scripts/check_compliance.py
```diff
@@ -614,6 +617,7 @@ def report_to_github(repo, pull_request, sha, suite, docs):
 
     return comment_count
 
+CURRENT_COMMIT_RANGE = "HEAD~1.."
```
ulfalizer: `TIP_COMMIT_ONLY(_RANGE)` or the like might be clearer. I misread `CURRENT_COMMIT_RANGE` as "currently active commit range" and was confused.

scripts/check_compliance.py
```diff
@@ -221,6 +221,9 @@ class Codeowners(ComplianceTest):
     def parse_codeowners(self, git_root, codeowners):
         all_files = []
         with open(codeowners, "r") as codeo:
+            if self.commit_range != CURRENT_COMMIT_RANGE:
+                logging.warning("Checking %s range against current %s file" %
+                                (self.commit_range, codeowners))
```
ulfalizer: Could add "...make sure it includes all commits that add files" maybe.

ulfalizer: Was this meant to be a `==` by the way?

If the commit range has just the tip commit, then that might miss checking other commits in the same PR that add files. I was assuming the message was there to warn against stuff like that, but I might be misunderstanding it.

Maybe this code could be moved to `Codeowners.run()`.

scripts/check_compliance.py
marc-hb: I did struggle to find a good name because it's a "single commit range" is a bit hard to express tersely. How about "HEAD_ONLY_RANGE"?

scripts/check_compliance.py
marc-hb: Most checks in this script allow testing any range, even past or future ranges that don't finish with ..HEAD. Let's say for instance that you have a very large series and you want to check just one early part of it.

However when looking at a past or future range, the Codeowners check stays stuck to the present CODEOWNERS file which will be incorrect if for instance you renamed a directory in the later part of your large series.

You're right: the current test is wrong but '==' is not the fix :-) It's wrong because it will print a **spurious** warning to any user checking HEAD~n.. where n>1

marc-hb: Codeowners.run() does look like a better place.

scripts/check_compliance.py
ulfalizer: Fine with me.

My check is "can this be read in more than one way?" `CURRENT_COMMIT_RANGE` can be read as `CURRENT_(COMMIT_RANGE)` and `(CURRENT_COMMIT)_RANGE`. `HEAD_ONLY_RANGE` is more difficult to read in two ways.

scripts/check_compliance.py
marc-h38: fix this

ulfalizer: Ah, yeah, should probably use the `CODEOWNERS` from the end of the range.

Even that isn't bulletproof, because it treats the entire range as a single commit w.r.t. `CODEOWNERS`. Could add a file in commit 1 and add it to `CODEOWNERS` in commit 5, and it wouldn't complain. It's a lot better though.

ulfalizer: Could parse the output of `git show <end of range>:CODEOWNERS` maybe.

marc-hb: > Even that isn't bulletproof, because it treats the entire range as a single commit 

BTW I have a work in progress to change that for checkpatch which does support checking commits one by one and should let remove this sort of approximation "for free". I digress.

carlescufi:
>Thanks for the prompt review. github doesn't let me request changes on my code to mark it as not ready doh! Can you do it for me?

You can use a DNM label for that

marc-hb:
This repo doesn't have any DNM label, as you saw I used a [DNM] prefix. Unlike the main one this repo seems to have only default github labels? Maybe another sign that this script should be in the main repo ;-)

marc-hb:
Pseudo-code of a hopefully correct implementation:
```python
if (git rev-list commit_range) != (git rev-list commit_range HEAD):
    # commit_range is old, doesn't end with HEAD
    logging.warning("checking old commits against present files and latest CODEOWNERS")
```
Think for instance of a file that was added in commit_range and then renamed later past it.
There are likely other ways this can go wrong; invoking this check on a commit_range that doesn't end with HEAD is most likely a mistake.

marc-hb:
If this warning ever gets implemented it won't look like this attempt, abandoning.

Commit fa8374b0a235 added a warning/TODO in the source code about this.

PR #26, opened 2019-03-04
check_compliance.py: Add test status reporting helpers

ulfalizer:
Add three helpers error(), skip(), and add_failure() to the
ComplianceTest class to signal problems running tests, skipped tests,
and test failures, respectively. These remove some boilerplate.

Derive the message passed to the Error, Skipped, and Failure
constructors from the test name.

Allow add_failure() to be called many times within the same test to
report multiple failures. This just concatenates the failure messages.

Also add some more information to some of the errors (and remove some
info from some, where things might be clear from the context).

Also turn two 'Error's in the CODEOWNERS and documentation tests into
'Failure's. 'Error' seems to be for reporting problems running the test,
and these are more like test failures.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -91,6 +91,59 @@ class ComplianceTest:
         """
         pass
 
+    def error(self, msg):
+        """
+        Signals a problem with running the test, with message 'msg'.
+
+        Raises an exception internally, so you do not need to put a 'return'
+        after error().
+
+        Any failures generated prior to the error() are included automatically
+        in the message. Usually, any failures would indicate problems with the
+        test code.
+        """
+        if self.case.result:
+            msg += "\n\nFailures before error: " + self.case.result._elem.text
+
+        self.case.result = Error(msg, "skipped")
+
+        raise EndTest
+
+    def skip(self, msg):
+        """
+        Signals that the test should be skipped, with message 'msg'.
+
+        Raises an exception internally, so you do not need to put a 'return'
+        after error().
+
+        Any failures generated prior to the skip() are included automatically
+        in the message. Usually, any failures would indicate problems with the
+        test code.
+        """
+        if self.case.result:
+            msg += "\n\nFailures before skip: " + self.case.result._elem.text
+
+        self.case.result = Skipped(msg, "skipped")
+
+        raise EndTest
+
+    def add_failure(self, msg):
+        """
+        Signals that the test failed, with message 'msg'. Can be called many
+        times within the same test to report multiple failures.
+        """
+        if not self.case.result:
+            # First reported failure
+            self.case.result = Failure(self._name + " issues", "failure")
+            self.case.result._elem.text = ""
+
+        # If there are multiple Failures, concatenate their messages
+        self.case.result._elem.text += msg + "\n\n"
+
+
+class EndTest(Exception):
+    "Raised by ComplianceTest.error()/skip() to end the test"
+
 
 class CheckPatch(ComplianceTest):
     """
@@ -106,7 +159,7 @@ class CheckPatch(ComplianceTest):
         checkpatch = os.path.join(self.zephyr_base or self.repo_path, 'scripts',
                                   'checkpatch.pl')
         if not os.path.exists(checkpatch):
-            self.case.result = Skipped("checkpatch script not found", "skipped")
+            self.skip(checkpatch + " not found")
 
         diff = subprocess.Popen(('git', 'diff', '%s' % (self.commit_range)),
                                 stdout=subprocess.PIPE)
@@ -116,10 +169,9 @@ class CheckPatch(ComplianceTest):
                                     stderr=subprocess.STDOUT, shell=True)
 
         except subprocess.CalledProcessError as ex:
-            match = re.search("([1-9][0-9]*) errors,", ex.output.decode('utf8'))
-            if match:
-                self.case.result = Failure("Checkpatch issues", "failure")
-                self.case.result._elem.text = (ex.output.decode('utf8'))
+            output = ex.output.decode("utf-8")
+            if re.search("[1-9][0-9]* errors,", output):
+                self.add_failure(output)
 
 
 class KconfigCheck(ComplianceTest):
@@ -134,15 +186,13 @@ class KconfigCheck(ComplianceTest):
         self.prepare()
 
         if not self.zephyr_base:
-            self.case.result = Skipped("Not a Zephyr tree", "skipped")
-            return
+            self.skip("Not a Zephyr tree (ZEPHYR_BASE unset)")
 
         # Put the Kconfiglib path first to make sure no local Kconfiglib version is
         # used
         kconfig_path = os.path.join(self.zephyr_base, "scripts", "kconfig")
         if not os.path.exists(kconfig_path):
-            self.case.result = Error("Can't find Kconfig", "error")
-            return
+            self.error(kconfig_path + " not found")
 
         sys.path.insert(0, kconfig_path)
         import kconfiglib
@@ -168,9 +218,7 @@ class KconfigCheck(ComplianceTest):
         try:
             kconf = kconfiglib.Kconfig()
         except kconfiglib.KconfigError as e:
-            self.case.result = Failure("error while parsing Kconfig files",
-                                       "failure")
-            self.case.result._elem.text = str(e)
+            self.add_failure(str(e))
             return
 
         #
@@ -180,12 +228,9 @@ class KconfigCheck(ComplianceTest):
         undef_ref_warnings = [warning for warning in kconf.warnings
                               if "undefined symbol" in warning]
 
-        # Generating multiple JUnit <failure>s would be neater, but Shippable only
-        # seems to display the first one
         if undef_ref_warnings:
-            self.case.result = Failure("undefined Kconfig symbols", "failure")
-            self.case.result._elem.text = "\n\n\n".join(undef_ref_warnings)
-            return
+            self.add_failure("Undefined Kconfig symbols:\n\n"
+                             + "\n\n\n".join(undef_ref_warnings))
 
         #
         # Check for stuff being added to the top-level menu
@@ -203,12 +248,11 @@ class KconfigCheck(ComplianceTest):
             node = node.next
 
         if n_top_items > max_top_items:
-            self.case.result = Failure("new entries in top menu", "failure")
-            self.case.result._elem.text = """
+            self.add_failure("""
 Expected no more than {} potentially visible items (items with prompts) in the
 top-level Kconfig menu, found {} items. If you're deliberately adding new
 entries, then bump the 'max_top_items' variable in {}.
-""".format(max_top_items, n_top_items, __file__)
+""".format(max_top_items, n_top_items, __file__))
 
 
 class Codeowners(ComplianceTest):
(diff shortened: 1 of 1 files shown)


ulfalizer:
Running on one hour of sleep (had an errand), so check carefully. :P

scripts/check_compliance.py
```diff
@@ -91,6 +91,44 @@ def run(self):
...
+        Signals that the test should be skipped, with message 'msg'.
+
+        Overrides fail() and error() if those have been called earlier.
+
+        Note: Does not raise an exception.
+        """
+        self.case.result = Skipped(self._name + " skipped", "skipped")
+        self.case.result._elem.text = msg
+
+    def fail(self, msg):
```
marc-hb: How about something like this:

```python

def addFailure(...): # or any other different looking name to highlight the difference with single-use .skip() and error()

if self.case.results and not isinstance(Failure):
   raise "something terrible happened"

if not self.case.results:
     self.case.results = Failure(...)

self.case.result._elem.text += msg + "\n\n"
```
"early return" coding style https://lwn.net/Articles/723489/

scripts/check_compliance.py
```diff
@@ -304,10 +335,7 @@ def run(self):
 
         if os.path.exists(self.DOCS_WARNING_FILE) and os.path.getsize(self.DOCS_WARNING_FILE) > 0:
             with open(self.DOCS_WARNING_FILE, "rb") as docs_warning:
-                log = docs_warning.read()
-
-                self.case.result = Error("Documentation Issues", "failure")
-                self.case.result._elem.text = log.decode('utf8')
+                self.fail(docs_warning.read().decode("utf-8"))
```
marc-hb: Could you use the opportunity to add a new DEFAULT_ENCODING = "utf-8" constant?

scripts/check_compliance.py
```diff
@@ -91,6 +91,44 @@ def run(self):
         """
         pass
 
+    def error(self, msg):
+        """
+        Signals a problem with running the test, with message 'msg'.
+
+        Overrides fail() and skip() if those have been called earlier.
+
+        Note: Does not raise an exception.
```
```python
def error(...):

if self.case.result:
  raise "Something went wrong"

File: scripts/check_compliance.py (hunk 1 above)
Comment (ulfalizer) [reply scripts/check_compliance.py]: The ability to override a failure with a later skipped, etc., is deliberate (and documented), for flexibility. Dunno if it's that useful in practice though.

Had this at first (`text` starts out as `None`), but those redundant `"\n\n"` for a single failure bothered me. :)

```python
if not isinstance(self.case.result, Failure):
    # If the result is not a Failure (usually because the result hasn't
    # been set yet), then make it one
    self.case.result = Failure(self._name + " issues", "failure")
    self.case.result._elem.text = ""

# If there are multiple Failures, concatenate their messages
self.case.result._elem.text += "\n\n" + msg
```

scripts/check_compliance.py
ulfalizer: I prefer the `"utf-8"` here. It'll never be anything else in practice, and probably won't be misspelled without raising an exception either.

ulfalizer:
Renamed `fail()` to `add_failure()`.

Kept the overriding functionality for now (and documented it for now). Not hardcore about it though. Could argue that it could hide issues in the code too...

scripts/check_compliance.py
marc-hb: Windows defaults to UTF-16 in a surprising number of places.
Exception raising assumes 100% test coverage, error handling has notoriously bad coverage - that's where many bugs go and hide.
Duplicated constants are just bad :-)
Anyway that was a bit off-topic, forget it.

ulfalizer: Won't get UTF-16 for those cases on Windows either.

Not always bad. `decode("utf-8")` gives you more information than `decode(DEFAULT_ENCODING)` a lot quicker, and isn't any more unsafe here in practice.

Personally, I'm big on shaving away as many lines as possible. Bugs hide in clutter. :)

scripts/check_compliance.py
marc-hb: > The ability to override a failure with a later skipped, etc., is deliberate (and documented), 

Wait... so now reviewers are supposed to look at not just code but also pydoc carefully added by authors? :-/

> for flexibility. Dunno if it's that useful in practice though.

Considering it silently drops information I looks pretty harmful to me... Have you spotted any actual usage?

How about logging.warning() then? It would not change the current behavior and just flag (very..) suspicious use cases.

> though it only shows up in compliance.xml

Errr.. then who cares?

ulfalizer: > Wait... so now reviewers are supposed to look at not just code but also pydoc carefully added by authors? :-/

Should always document functions IMO (and read docs). Often difficult to know what's intentional and what isn't otherwise.

> Considering it silently drops information I looks pretty harmful to me... Have you spotted any actual usage?

Nope.

I wonder if it's that bad in practice though. Feels like the worst thing that could happen would be that someone would have to run CI twice. Hard to even come up with cases where that'd happen, and it'd probably be obvious.

If someone wants to say "oh wait, skip this test" after adding some failures, then let them, imo. Keeps the code short and simple.

> 
> How about logging.warning() then? It would not change the current behavior and just flag (very..) suspicious use cases.

Don't think it'll ever be a problem in practice. Better to keep it short.

ulfalizer: > > though it only shows up in compliance.xml
> 
> Errr.. then who cares?

My OCD. Changed.

marc-hb: > If someone wants to say "oh wait, skip this test" after adding some failures, then let them, imo

I can't think of why you'd want to skip() *after* adding failures or errors - or any other similar use case.

+1 except for this.

> Feels like the worst thing that could happen would be that someone would have to run CI twice. 

Sorry I don't get this.

ulfalizer:
Hmm... maybe `skip()` and `error()` could (always) raise an exception. That way, it'd be impossible to forget a `return`, and it simplifies the callers a bit.

ulfalizer:
Will do that. Could include any `Failure` message at that point too, as a debugging aid, to be a bit more diplomatic ("... Warning: failures were generated before the Error/Skip: <message>").

Lack of sleep makes me annoyingly cranky...

ulfalizer:
@marc-hb 
Pushed a ew version that aborts the test immediately on `error()` and `skip()`. If there are any reported failures at that point, they get included in the message.

marc-hb:
I just tried this patch. Before:
```
WARNING : Skipped Kconfig, Not a Zephyr tree
WARNING : Skipped License, scancode-toolkit not installed
```
After:
```
WARNING : Skipped Kconfig, Kconfig skipped
WARNING : Skipped License, License skipped
```
I admittedly keep getting lost in the various msg fields... is this because of the new exceptions? In any case it's much less useful sorry :-(

ulfalizer:
@marc-hb 
Fixed. Diff:

```diff
diff --git a/scripts/check_compliance.py b/scripts/check_compliance.py
index d4eaed2..4b9e0d8 100755
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -124,8 +124,7 @@ class ComplianceTest:
         if self.case.result:
             msg += "\n\nFailures before skip: " + self.case.result._elem.text
 
-        self.case.result = Skipped(self._name + " skipped", "skipped")
-        self.case.result._elem.text = msg
+        self.case.result = Skipped(msg, "skipped")
 
         raise EndTest
```

Maybe setting `self.case.result._elem.text` isn't helpful for `Skipped`. Can't remember how it shows up in the Shippable interface.

scripts/check_compliance.py
```diff
@@ -702,15 +737,17 @@ def main():
...
+
+        try:
             test.run()
-            suite.add_testcase(test.case)
+        except EndTest:
+            pass
+        suite.add_testcase(test.case)
 
     xml = JUnitXml()
     xml.add_testsuite(suite)
```
marc-hb: @line 779 would you mind fixing this related and misleading print:

```patch
- print("{} Errors found".format(errors))
+ print("{} checks failed".format(errors))
```
It's really *not* a number of errors, that just really confused me.

ulfalizer: Fixed

scripts/check_compliance.py
```diff
@@ -134,15 +187,13 @@ def run(self):
...
-            return
+            self.skip("Not a Zephyr tree (ZEPHYR_BASE unset)")
 
         # Put the Kconfiglib path first to make sure no local Kconfiglib version is
         # used
         kconfig_path = os.path.join(self.zephyr_base, "scripts", "kconfig")
         if not os.path.exists(kconfig_path):
-            self.case.result = Error("Can't find Kconfig", "error")
-            return
+            self.error(kconfig_path + " not found")
```
marc-hb: The new code changes the behavior here: it moves the text from the Error(msg=text, ...) constructor to ._elem.text. Is this intended and will github/JUnit be happy with this?

(basically the same discussion than the Skip() change you just did)

scripts/check_compliance.py
```diff
@@ -304,10 +347,7 @@ def run(self):
 
         if os.path.exists(self.DOCS_WARNING_FILE) and os.path.getsize(self.DOCS_WARNING_FILE) > 0:
             with open(self.DOCS_WARNING_FILE, "rb") as docs_warning:
-                log = docs_warning.read()
-
-                self.case.result = Error("Documentation Issues", "failure")
-                self.case.result._elem.text = log.decode('utf8')
+                self.add_failure(docs_warning.read().decode("utf-8"))
```
marc-hb: Hey maybe this change from Error to Failure will provide some more transparency? see discussion at https://lists.zephyrproject.org/g/devel/topic/30039752

Found this interesting bit:
http://doc.scalatest.org/1.0/org/scalatest/junit/JUnitTestFailedError.html
"In JUnit 4, AssertionFailedError was made to extend java.lang.AssertionError, and the distinction between failures and errors was essentially dropped. However, some tools that integrate with JUnit carry on this distinction,..."

scripts/check_compliance.py
ulfalizer: Think it'll be fine. I think the constructor argument is more like a short summary/title, which is always shown, while the `_text` stuff is the body text in e.g. the GitHub comment and on Shippable.

ulfalizer: Oh... wait, this does put it in the `_text`. Confused by my own helpers.

Yeah, maybe it should be moved. I'll do a test commit to check how it renders.

ulfalizer: Seems you get neither the constructor argument nor the body text anywhere, on GitHub at least. Just a "Error during verification, please report!".

I dropped a random `self.error("yada yada")` at the start of a test.

ulfalizer: I changed `error()` to work the same as `skip()` (will get a really long "title" if there are any failures prior, but `skip()` has that problem too).

Looks like the `error()` message might not show up anywhere on GitHub anyway (dunno about Shippable), so that might need to be fixed separately.

Would really help to be able to check Shippable output in these PRs.

scripts/check_compliance.py
ulfalizer: Yeah, especially if the `Error` message is just dropped by default.

ulfalizer:
@marc-hb 
Any more objections? Could approve otherwise.

Would be nice to get this in so that I can put https://github.com/zephyrproject-rtos/ci-tools/pull/23 and https://github.com/zephyrproject-rtos/ci-tools/pull/21 on top of it.

PR #27, opened 2019-03-05
scripts: check_compliance: Check for kconfig variables declared

sigvartmh:
Add support for a single file as source of truth for environmental
variables declared for Kconfig in Cmake. It's coupled with a change in
the cmake build system where you declare the environmental variables for
Kconfig in a single text file. This will give you an error if the script
does not set the required variables and will list which variables are
missing.

This is coupled with: https://github.com/zephyrproject-rtos/zephyr/pull/14072

Signed-off-by: Sigvart Hovland <sigvart.m@gmail.com>

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -156,6 +156,9 @@ class KconfigCheck(ComplianceTest):
         os.environ["BOARD_DIR"] = "boards/*/*"
         os.environ["ARCH"] = "*"
         os.environ["PROJECT_BINARY_DIR"] = tempfile.gettempdir()
+        os.environ["PYTHON_EXECUTABLE"] = os.path.normpath(sys.executable)
+        os.environ["KCONFIG_CONFIG"] = "dummy"
+        os.environ["KERNELVERSION"] = "dummy"
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 
         # For multi repo support
@@ -164,6 +167,16 @@ class KconfigCheck(ComplianceTest):
         # Enable strict Kconfig mode in Kconfiglib, which assumes there's just a
         # single Kconfig tree and warns for all references to undefined symbols
         os.environ["KCONFIG_STRICT"] = "y"
+
+        kconfig_env_var_list_path = os.path.join(self.zephyr_base, "cmake/kconfig_env_vars.txt")
+        kconfig_env_vars = set(open(kconfig_env_var_list_path).read().splitlines())
+
+        missing_vars = kconfig_env_vars.difference(os.environ) 
+        if missing_vars:
+            self.case.result = Failure("error environmental variables {1} listed in {0} "
+                                       "are not declared in check_compliance.py".format(
+                                       kconfig_env_var_list_path, missing_vars), "failure")
+            return
 
         try:
             kconf = kconfiglib.Kconfig()


sigvartmh:
@SebastianBoe

scripts/check_compliance.py
```diff
@@ -165,6 +167,23 @@ def run(self):
         # single Kconfig tree and warns for all references to undefined symbols
         os.environ["KCONFIG_STRICT"] = "y"
 
+        kconfig_env_var_list_path = os.path.join(self.zephyr_base,
+                                    os.path.normpath("cmake/kconfig_env_vars.txt"))
+        # Skipping last new line which gives empty string
+        kconfig_env_var_list = open(kconfig_env_var_list_path).read().split("\n")[:-1]
```
ulfalizer: Can use `readlines()`.

scripts/check_compliance.py
```diff
@@ -165,6 +167,23 @@ def run(self):
         # single Kconfig tree and warns for all references to undefined symbols
         os.environ["KCONFIG_STRICT"] = "y"
 
+        kconfig_env_var_list_path = os.path.join(self.zephyr_base,
+                                    os.path.normpath("cmake/kconfig_env_vars.txt"))
```
ulfalizer: Can drop the `os.path.normpath()`. Will be a no-op for `"cmake/kconfig_env_vars.txt"`.

scripts/check_compliance.py
```diff
@@ -165,6 +167,23 @@ def run(self):
...
+        # Skipping last new line which gives empty string
+        kconfig_env_var_list = open(kconfig_env_var_list_path).read().split("\n")[:-1]
+
+        # Faster to check for all than iterate over every variable
+        if not all(env_var in os.environ for env_var in kconfig_env_var_list):
+            missing_vars=[]
+            for env_var in kconfig_env_var_list:
+                if env_var not in os.environ.keys():
```
ulfalizer: Can do `if env_var not in os.environ:`. Iteration and membership testing always uses the keys.

scripts/check_compliance.py
```diff
@@ -165,6 +167,23 @@ def run(self):
         # single Kconfig tree and warns for all references to undefined symbols
         os.environ["KCONFIG_STRICT"] = "y"
 
+        kconfig_env_var_list_path = os.path.join(self.zephyr_base,
+                                    os.path.normpath("cmake/kconfig_env_vars.txt"))
+        # Skipping last new line which gives empty string
+        kconfig_env_var_list = open(kconfig_env_var_list_path).read().split("\n")[:-1]
+
+        # Faster to check for all than iterate over every variable
+        if not all(env_var in os.environ for env_var in kconfig_env_var_list):
```
ulfalizer: Sets are pretty handy here:

```python
kconfig_env_vars = set(open(...).readlines())

missing_vars = kconfig_env_vars.difference(os.environ)
if missing_vars:
    ...
```

scripts/check_compliance.py
```diff
@@ -165,6 +167,23 @@ def run(self):
...
+            missing_vars=[]
```
ulfalizer: Missing spaces around `=`.

ulfalizer:
One thing that limits the usefulness is that the list really only checks that those variables are set, and not their values.

Have you run into cases where someone forgot to set an environment variable and it lead to silent errors?

scripts/check_compliance.py
sigvartmh: but doesn't `readlines()` append the `\n` to the string?

scripts/check_compliance.py
sigvartmh: Yeah that's what the idea was but I think `os.path.join` will fix it anyways so I guess i'll remove it

scripts/check_compliance.py
ulfalizer: Ah, right, had forgotten that. `.read().splitlines()` seems to work.

scripts/check_compliance.py
sigvartmh: That's a great idea I'll do that instead :) Thanks for the tip

sigvartmh:
> 
> 
> One thing that limits the usefulness is that the list only checks that those variables are set, and not their values.
> 
> Have you run into cases where someone forgot to set an environment variable and it lead to silent errors?

I can add an `if empty` check for the forgot case but I can't find a way to check the validity of what is set. Originally I wanted to do a check if the values are correct but I couldn't find a way to get the actual generated values as this script doesn't invoke CMake.

scripts/check_compliance.py
sigvartmh: yeah, that's better than manually trying to split for `\n`

sigvartmh: I'll add that instead

ulfalizer:
> > One thing that limits the usefulness is that the list only checks that those variables are set, and not their values.
> > Have you run into cases where someone forgot to set an environment variable and it lead to silent errors?
> 
> I can add an `if empty` check for the forgot case but I can't find a way to check the validity of what is set. Originally I wanted to do a check if the values are correct but I couldn't find a way to get the actual generated values as this script doesn't invoke CMake.

Thinking more along the lines of what happens now if you forget to set one of the environment variables in the list. Do you ever get a silent failure, or do you always get e.g. a crash (that hopefully points you to the source of the problem)? I should know this, but I'm not sure for all of them. :)

If there are no silent failures, then maybe parsing the list of environment variables adds more complexity to the code than it helps out with debugging.

sigvartmh:
> > > One thing that limits the usefulness is that the list only checks that those variables are set, and not their values.
> > > Have you run into cases where someone forgot to set an environment variable and it lead to silent errors?
> > 
> > 
> > I can add an `if empty` check for the forgot case but I can't find a way to check the validity of what is set. Originally I wanted to do a check if the values are correct but I couldn't find a way to get the actual generated values as this script doesn't invoke CMake.
> 
> Thinking more along the lines of what happens now if you forget to set one of the environment variables in the list. Do you ever get a silent failure, or do you always get e.g. a crash (that hopefully points you to the source of the problem)? I should know this, but I'm not sure for all of them. :)
> 
> If there are no silent failures, then maybe parsing the list of environment variables adds more complexity to the code than it helps out with debugging.

scripts/check_compliance.py
```diff
@@ -156,6 +156,9 @@ def run(self):
         os.environ["BOARD_DIR"] = "boards/*/*"
         os.environ["ARCH"] = "*"
         os.environ["PROJECT_BINARY_DIR"] = tempfile.gettempdir()
+        os.environ["PYTHON_EXECUTABLE"] = os.path.normpath(sys.executable)
```
sigvartmh: This is actually not needed by anything but is set to make the test which checks which env variables are needed to be set by `kconfig.cmake` so that when external/third-party scripts want to call cmake they can use the same lookup list as this test is using `kconfig_env_vars.txt`at least that's the idea behind of the change.

scripts/check_compliance.py
sigvartmh: Done

scripts/check_compliance.py
sigvartmh: Changed to set instead

scripts/check_compliance.py
sigvartmh: Sets was much nicer thanks once again 👍

scripts/check_compliance.py
sigvartmh: removed

scripts/check_compliance.py
sigvartmh: fixed

ulfalizer:
> By setting do you mean defining the environmental variable but not assigning it an value? If you don't set it you get an failure, if you set it but not to anything just defining it then it's a bit unknown what would happen since not all env variables are used by the script. Those which are required by the script will end in a failure other would end in the compliance test failing.

I mean if you don't set it.

What the PR adds is a check that all required environment variables are set (to anything). If you get an error if you forget to set any of them, then that in itself might be good enough though.

Could still document all the environment variables somewhere.

sigvartmh:
> I mean if you don't set it.
> 
> What the PR adds is a check that all required environment variables are set (to anything). If you get an error if you forget to set any of them, then that in itself might be good enough though.
> 
> Could still document all the environment variables somewhere.

I hear what you say and of course early failures/errors over silently fail is better, but the motivation behind this PR was the behaviour experienced by not setting PROJECT_BINARY_DIR variable correctly in this script outputs this error:

```
Kconfig.zephyr:41: '/Kconfig.modules' does not exist. Perhaps the $srctree
environment variable (set to '/var/lib/shippable/build/IN/main_repo/zephyr') is
set incorrectly. Note that the current value of $srctree is saved when the
Kconfig instance is created (for consistency and to cleanly separate instances).
```

Which kind of kicks you off in the wrong direction. So the idea was to atleast try to create a better error message for this case and then hopefully catch other future cases, but I do see your point that it may confuse and make people set envvars making something fail silently :/ 

I'm a bit unsure if it adds more value than it removes.

ulfalizer:
Ah, yeah, that Kconfiglib error message could be improved. `srctree` being the wrong value probably isn't the real issue that often.

Would it help if it said this instead?

```
Kconfig.zephyr:41: '/Kconfig.modules' does not exist (in line 'source "$(PROJECT_BINARY_DIR)/Kconfig.modules"'). Check that environment variables are set correctly.
```

ulfalizer:
Here's a PR that improves the error message for missing files in `source`: https://github.com/zephyrproject-rtos/zephyr/pull/14085

Tried to add you as a reviewer, but not possible, for some reason.

Did you spot any other bad ones?

scripts/check_compliance.py
```diff
@@ -165,6 +168,16 @@ def run(self):
         # single Kconfig tree and warns for all references to undefined symbols
         os.environ["KCONFIG_STRICT"] = "y"
 
+        kconfig_env_var_list_path = os.path.join(self.zephyr_base, "cmake/kconfig_env_vars.txt")
+        kconfig_env_vars = set(open(kconfig_env_var_list_path).read().splitlines())
```
SebastianBoe: The file should be closed, is it? I can't tell.

sigvartmh: yup it should probably do a close call as I'm not using `with` I'm too used to using `with` keyword which does this for you. Not sure where the close should be put I guess after `read`

ulfalizer: In CPython (the default Python implementation), files will be closed immediately when the last reference to them goes away (which is after that line).

The reason it's still a good idea (but not always super important) to use `with open(...) ...` is that some other Python implementations don't use simple reference counting, like PyPy.

ulfalizer: `with` is used like this by the way:

```python
with open(...) as f:
    f.read(), etc.
```

sigvartmh:
> Here's a PR that improves the error message for missing files in `source`: [zephyrproject-rtos/zephyr#14085](https://github.com/zephyrproject-rtos/zephyr/pull/14085)
> 
> Tried to add you as a reviewer, but not possible, for some reason.
> 
> Did you spot any other bad ones?

Sorry been out for a while, can't remember any more atm. I'll give you a ping if I find some :) I guess this change is not really needed then?

ulfalizer:
Closing for now.

PR #31, opened 2019-03-08
check_compliance.py: Break Kconfig tests into several functions

ulfalizer:
This can be done in a neat way with the new
`ComplianceTest.error()`/`skip()`/`add_failure()` helpers.

Have a separate function for parsing the `Kconfigs`, and a separate
function for each test. Only parse once and reuse the `kconfiglib.Kconfig`
object between tests, to save a little bit of time.

Also allow tests to raise `EndTest` themselves wherever it's handy, and
document it.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -142,7 +142,12 @@ class ComplianceTest:
 
 
 class EndTest(Exception):
-    "Raised by ComplianceTest.error()/skip() to end the test"
+    """
+    Raised by ComplianceTest.error()/skip() to end the test.
+
+    Tests can raise EndTest themselves to immediately end the test, e.g. from
+    within a nested function call.
+    """
 
 
 class CheckPatch(ComplianceTest):
@@ -185,6 +190,16 @@ class KconfigCheck(ComplianceTest):
     def run(self):
         self.prepare()
 
+        kconf = self.parse_kconfig()
+
+        self.check_no_undef_within_kconfig(kconf)
+        self.check_top_menu_not_too_long(kconf)
+
+    def parse_kconfig(self):
+        """
+        Returns a kconfiglib.Kconfig object for the Kconfig files. We reuse
+        this object for all tests to avoid having to reparse for each test.
+        """
         if not self.zephyr_base:
             self.skip("Not a Zephyr tree (ZEPHYR_BASE unset)")
 
@@ -216,15 +231,16 @@ class KconfigCheck(ComplianceTest):
         os.environ["KCONFIG_STRICT"] = "y"
 
         try:
-            kconf = kconfiglib.Kconfig()
+            return kconfiglib.Kconfig()
         except kconfiglib.KconfigError as e:
             self.add_failure(str(e))
-            return
+            raise EndTest
 
-        #
-        # Look for undefined symbols
-        #
-
+    def check_no_undef_within_kconfig(self, kconf):
+        """
+        Checks that there are no references to undefined Kconfig symbols within
+        the Kconfig files
+        """
         undef_ref_warnings = [warning for warning in kconf.warnings
                               if "undefined symbol" in warning]
 
@@ -232,10 +248,11 @@ class KconfigCheck(ComplianceTest):
             self.add_failure("Undefined Kconfig symbols:\n\n"
                              + "\n\n\n".join(undef_ref_warnings))
 
-        #
-        # Check for stuff being added to the top-level menu
-        #
-
+    def check_top_menu_not_too_long(self, kconf):
+        """
+        Checks that there aren't too many items in the top-level menu (which
+        might be a sign that stuff accidentally got added there)
+        """
         max_top_items = 50
 
         n_top_items = 0


scripts/check_compliance.py
```diff
@@ -216,26 +231,28 @@ def run(self):
...
-            kconf = kconfiglib.Kconfig()
+            return kconfiglib.Kconfig()
         except kconfiglib.KconfigError as e:
             self.add_failure(str(e))
-            return
-
-        #
-        # Look for undefined symbols
-        #
+            raise EndTest
```
```python
self.error(str(e))
```
?

scripts/check_compliance.py
```diff
@@ -216,26 +231,28 @@ def run(self):
...
 
         if undef_ref_warnings:
             self.add_failure("Undefined Kconfig symbols:\n\n"
                              + "\n\n\n".join(undef_ref_warnings))
 
-        #
-        # Check for stuff being added to the top-level menu
-        #
-
+    def check_n_top_menu_items(self, kconf):
```
marc-hb: I'd like to have either "max" or "too many" or "too_long_top_menu" or something like it in the function name.

scripts/check_compliance.py
```diff
@@ -216,26 +231,28 @@ def run(self):
...
         except kconfiglib.KconfigError as e:
             self.add_failure(str(e))
-            return
-
-        #
-        # Look for undefined symbols
-        #
+            raise EndTest
 
+    def check_undef_within_kconfig(self, kconf):
```
marc-hb: Naive question: would it be possible to just make the actual build fail in this case? I mean within kconfiglib.py in the main zephyr repo.
AFAIK check_compliance.py is for either "soft" rules that need to be overridden from time to time, or for checks that are just too difficult or too slow to implement inside the main build itself. In which category is this one?

scripts/check_compliance.py
ulfalizer: Problem is `error()` messages aren't shown, and this one will point to the file/line of the problem, so it's pretty important. Dunno about shippable though.

Could view it as a Kconfig pre-test too. Do the files parse? :P

scripts/check_compliance.py
ulfalizer: One thing that's different is that you get a pruned version of the Kconfig tree when running samples and the like (via `source`s that have environment variables in them). This test runs with everything included.

scripts/check_compliance.py
ulfalizer: The message might also be pretty long if it's from Kconfiglib, so might be nice to have it in the body.

scripts/check_compliance.py
ulfalizer: Renamed it to `check_top_menu_not_too_long()`.

scripts/check_compliance.py
marc-hb: > Problem is error() messages aren't shown

Do you mean on github? I see them on the command line.

scripts/check_compliance.py
ulfalizer: Oh, I thought this was for the too-long-menu-test at first, where it's not as essential to see the entire Kconfig tree (but still nice).

For the undefined-within-Kconfig test, it's essential to see the entire Kconfig tree. Otherwise, you'll end up with false positives for symbols that are defined in files that aren't parsed.

You get the same thing in the Linux kernel too. A generic symbol that shows up during an X86 build might depend on `!SOME_ARM_SPECIFIC_THING`, which ends up being undefined. That's fine, because undefined symbols evaluate to `n`.

scripts/check_compliance.py
ulfalizer: Yup, on GitHub (iirc it just says "error blah blah please report this"), and maybe Shippable.

Errors in the Kconfig files aren't a problem with the test itself, so I think the `add_failure()` kinda makes sense there.

scripts/check_compliance.py
marc-hb: >  ... pruned version of the Kconfig tree when running samples and the like (via sources that have environment variables in them).

Thanks I wasn't aware. Can you point at one specific example?

> You get the same thing in the Linux kernel too

I'm confused. When I run `make menuconfig` in the Linux kernel, I (think I) can search for any obscure sub-sub-sub option, even one that requires a cascade of other dependencies and conditions before it becomes available and shows up in the menus (that's probably the most frequent reason I use kconfig's help: find stuff). I don't see how kconfig could do this without parsing the entire Kconfig tree.

ulfalizer: Search for `$(ARCH)`, `$(SOC_DIR)`, and `$(BOARD_DIR)` in `Kconfig.zephyr` (that's the syntax for referencing environment variables). Those get set before Kconfig runs, in `cmake/kconfig.cmake`.

For Linux, there's `source "arch/$(SRCARCH)/Kconfig"` in `arch/Kconfig`, which includes e.g. `arch/arm/Kconfig`. Try searching for symbols defined in there while building for x86, and you'll see that they're missing.

marc-hb: OK, so you can get the same thing in the Linux kernel too, however if excluding Kconfig files there depends only on SRCARCH (and no SOC_DIR, BOARD_DIR, GPIO42_DIR,...) then it's very unlikely a developer will submit a Kconfig change without testing it with the corresponding ARCH.

Now for fun I purposely broke a couple Kconfig files in the Linux kernel and found that its (original, C and unrelated) kconfig implementation generally doesn't seem to care about undefined symbols and other typos - not even mistakes in the Kconfig files that it does parse. Pretty much anything seems to go, even inverted ranges. I haven't found any "checkconfig" feature in the Linux kernel either so I'm afraid the Linux part of the discussion above was basically entirely irrelevant :-/

marc-hb: Leaving Linux aside and rephrasing with a bit more knowledge my initial suggestion at the very top of this thread: why not move this `check_compliance.py#KconfigCheck()` code to `ZEPHYR_BASE/script/kconfig/checkconfig2.py` and ask `cmake` to run it every time anyone builds anything? It seems to take a split second to run and anyone making kconfig changes would most certainly appreciate their mistakes getting caught locally long before CI runs and without the round trips.

ulfalizer: Adds a bit of overhead that people that reconfigure often might not like (1-2 secs on my Core i7 2600k, though it gets faster once the files are cached). One reason I think people might dismiss Kconfiglib is that they'll assume it'll be too slow since it's doing parsing in Python, so I've gotten kinda obsessed with having it speedy (by Python standards). :)

Maybe it can get annoying during development too, like being forced to constantly run a linter on WIP stuff.

I don't rule it out in the future though. None of my code is ever "done". Stuff can always be improved, and I tend to go back to stuff.

ulfalizer: And yeah, the C tools are actually kinda crap in many ways, and completely unreadable (which led to the crappiness, because people are afraid of touching the code and breaking something).

marc-hb: > And yeah, the C tools are actually kinda crap in many ways, and completely unreadable (which led to the crappiness, because people are afraid of touching the code and breaking something).

Small world: you just reminded me of https://lkml.org/lkml/2018/3/1/1191 "Add Kconfig unit tests" and now the name of one of the reviewers seems familiar :-D

ulfalizer: Yeah, I've done some cleanups/added some comments to the C tools too. Working on that code always feels a bit masochistic though. :)

ulfalizer: Any showstoppers? Could approve otherwise.

PR #33, opened 2019-03-12
scripts: check_compliance: fix for multi-image

sigvartmh:
Check_compliance fails for multi-image this should fix this. Should not be merged before https://github.com/zephyrproject-rtos/zephyr/pull/13672 is in.

Signed-off-by: Sigvart Hovland <sigvart.m@gmail.com>

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -205,6 +205,7 @@ class KconfigCheck(ComplianceTest):
         os.environ["ARCH_DIR"] = "arch/"
         os.environ["BOARD_DIR"] = "boards/*/*"
         os.environ["ARCH"] = "*"
+        os.environ["CMAKE_BINARY_DIR"] = tempfile.gettempdir()
         os.environ["PROJECT_BINARY_DIR"] = tempfile.gettempdir()
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 


scripts/check_compliance.py
```diff
@@ -205,6 +205,7 @@ def run(self):
         os.environ["ARCH_DIR"] = "arch/"
         os.environ["BOARD_DIR"] = "boards/*/*"
         os.environ["ARCH"] = "*"
+        os.environ["CMAKE_BINARY_DIR"] = tempfile.gettempdir()
```
marc-hb: Are PROJECT_BINARY_DIR and CMAKE_BINARY_DIR also the same when performing a regular build? If yes then why do both need to be manually configured? Seems error prone.

marc-hb: Also not critical and not something you did but it would be nice to:
- use a top-level directory "zephyr-tmp-something" instead of dumping everything at the /tmp top level. This would tell a bit where these files come from and simplify cleanup.
- have concurrent instances of check_compliance.py not using the same files, that's unlikely to work.

ulfalizer: I don't know how all of this stuff fits together yet, but one thing I think is a bad idea is to require environment variables to be set manually before building. Ideally, you should just be able to clone + (maybe) install some dependencies + run a single command to build (or maybe two).

Zephyr already requires environment variables to be set, and IMO we should get rid of it. The thing Contiki does [here](https://github.com/contiki-os/contiki/blob/master/examples/collect/Makefile#L5) is what I think is the good way to do it. It's about having good defaults, and only having to mess with environment variables if those won't work.

If this is complete unrelated, then please ignore it. Just thought I'd say it in case it is related. ;)

sigvartmh: > Are PROJECT_BINARY_DIR and CMAKE_BINARY_DIR also the same when performing a regular build? If yes then why do both need to be manually configured? Seems error prone.

For multi image no.

marc-hb: By "regular" I didn't mean "single image", I meant "not check_compliance.py". Sorry for the confusion.

So if the two DIRs are different for a non-check_compliance build, multi image build, is it OK for them to be the same for a check_compliance build?

More generally and importantly, what error and problem does this solve? The commit message must summarize the issue and solution and not just point at the giantic multi-image PR.

mbolivar: > Zephyr already requires environment variables to be set, and IMO we should get rid of it. 

That's what we're trying to do with west build, FWIW. The toolchain variables are what is left as far as I know. I'm working on eliminating the need for `west build` callers to set those as well.

sigvartmh:
Not going to maintain this PR anymore

PR #34, opened 2019-03-17
kconfig: generate Kconfig.modules from west list

nashif:
Get a list of modules and put them in a file that is parsed by Kconfig.

This is needed for verifying multi-repo configuration where some Kconfig
options are only defined in module git repo but not in the main tree.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -19,6 +19,7 @@ import json
 import tempfile
 from colorama import Fore, Back, Style
 import glob
+import yaml
 
 logger = None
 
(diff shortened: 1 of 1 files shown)


nashif:
> Issue #1: Should this be common code in CI and in regular builds?

where else is this needed? Why would you need this in regular builds?

scripts/check_compliance.py
```diff
@@ -47,6 +48,29 @@ def get_shas(refspec):
...
+    Get a list of modules and put them in a file that is parsed by Kconfig
+
+    This is needed to complete Kconfig sanity tests.
+
+    """
+    modules = sh.west("list", "--format={posixpath}", **sh_special_args)
+    with open(modules_file, 'w') as km:
+        for m in modules:
+            module_path = m.strip()
+            module_file = "{}/zephyr/module.yml".format(module_path)
```
carlescufi: Please use `os.path.join`. Windows paths are different

scripts/check_compliance.py
```diff
@@ -224,7 +248,7 @@ def parse_kconfig(self):
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 
         # For multi repo support
-        open(os.path.join(tempfile.gettempdir(), "Kconfig.modules"), 'a').close()
+        get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
```
carlescufi: How does the root Kconfig.zephyr file find the location of this tempfile?

carlescufi:
> > Issue #1: Should this be common code in CI and in regular builds?
> 
> where else is this needed? Why would you need this in regular builds?

I mean that we now have 2 implementations of this. One in CMake and this one in Python. Would it be better to move this Python code to the zephyr repo and reuse it here? Then we implement it only once, in Python, and remove the current CMake implementation

scripts/check_compliance.py
nashif: look a few lines up, it is in 

`$(PROJECT_BINARY_DIR)/Kconfig.modules` which translates to 

`os.path.join(tempfile.gettempdir(), "Kconfig.modules")`

nashif:
> I mean that we now have 2 implementations of this. One in CMake and this one in Python. Would it be better to move this Python code to the zephyr repo and reuse it here? Then we implement it only once, in Python, and remove the current CMake implementation

We could, but I would prefer for this to be implemented in the Zephyr tree, get finalised and tested then we can look into how to use it in CI, probably something for after 1.14. This is now resolve a CI issue an can always be optimised when we have a better way for doing this using code from the tree.

scripts/check_compliance.py
carlescufi: Oh right. I thought that gettempdir() returned a different one every time.

marc-hb:
Fixed crash in PR #36

PR #36, opened 2019-03-18
check_compliance: don't crash when west's missing, just skip Kconfig

marc-hb:
I can't find any reason to miss on all the (other) tests when only
Kconfig requires west (since very recent commit 8820c86161ec)

Instead of crashing with a stack trace this prints this:

Running checkpatch tests...
Running Kconfig tests...
Running Codeowners tests...
Running Documentation tests...
Running Gitlint tests...
Running License tests...
Running Identity/Emails tests...
WARNING : Skipped Kconfig, Command not found: west
WARNING : Skipped License, scancode-toolkit not installed

Complete results in compliance.xml

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -48,14 +48,8 @@ def get_shas(refspec):
                       refspec, **sh_special_args).split()
     return sha_list
 
-def get_modules(modules_file):
-    """
-    Get a list of modules and put them in a file that is parsed by Kconfig
 
-    This is needed to complete Kconfig sanity tests.
-
-    """
-    modules = sh.west("list", "--format={posixpath}", **sh_special_args)
+def create_modules_file(modules, modules_file):
     with open(modules_file, 'w') as km:
         for m in modules:
             module_path = m.strip()
@@ -223,6 +217,23 @@ class KconfigCheck(ComplianceTest):
         self.check_no_undef_within_kconfig(kconf)
         self.check_top_menu_not_too_long(kconf)
 
+    def get_modules(self, modules_file):
+        """
+        Get a list of modules and put them in a file that is parsed by
+        Kconfig
+
+        This is needed to complete Kconfig sanity tests.
+
+        """
+        try:
+            modules = sh.west("list", "--format={posixpath}", **sh_special_args)
+        except sh.CommandNotFound as e:
+            self.skip("Command not found: " + str(e))
+        except sh.ErrorReturnCode as e:
+            self.skip(e)
+
+        create_modules_file(modules, modules_file)
+
     def parse_kconfig(self):
         """
         Returns a kconfiglib.Kconfig object for the Kconfig files. We reuse
@@ -252,7 +263,7 @@ class KconfigCheck(ComplianceTest):
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 
         # For multi repo support
-        get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
+        self.get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
 
         # Enable strict Kconfig mode in Kconfiglib, which assumes there's just a
         # single Kconfig tree and warns for all references to undefined symbols


scripts/check_compliance.py
```diff
@@ -252,7 +252,10 @@ def parse_kconfig(self):
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 
         # For multi repo support
-        get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
+        try:
+            get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
+        except sh.CommandNotFound as e:
+            self.skip("Command not found: " + str(e))
```
ulfalizer: Could add a comment along the lines of `# Skip test if west isn't installed` maybe, to make it clearer what it's about.

marc-hb: Agree it wouldn't hurt... on the other hand I suspect no one will look at these lines unless they saw the "Skipped because west's missing" message first. For the rare person who didn't see that message, a quick look at the short get_modules() function shows only one invocation of sh: `sh.west`

I will add that comment if this quick fix requires more time because other, more important problems are found.

marc-hb:
Needs more work: this doesn't deal with the case when `west` is installed but the current project is not a west project.

ulfalizer:
> Needs more work: this doesn't deal with the case when `west` is installed but the current project is not a west project.

Thought that had been fixed. IMO, whether you have `west` installed or not shouldn't matter if you're in a non-west Zephyr directory. The real bug is there.

marc-hb:
> Thought that had been fixed.

Are you thinking of this? https://github.com/zephyrproject-rtos/zephyr/issues/14177 It's a different issue

> whether you have west installed or not shouldn't matter if you're in a non-west Zephyr directory. 

These two different problems are reported by two different exceptions, which I think is OK. Just need to catch both.

ulfalizer:
> > Thought that had been fixed.
> 
> Are you thinking of this? [zephyrproject-rtos/zephyr#14177](https://github.com/zephyrproject-rtos/zephyr/issues/14177) It's a different issue

Right, that was the one. Nice that it got fixed!

Never mind the rambling.

scripts/check_compliance.py
marc-hb: In the second and newer version `sh.west` is now on the line literally before `sh.CommandNotFound`.

scripts/check_compliance.py
```diff
@@ -223,6 +217,23 @@ def run(self):
         self.check_no_undef_within_kconfig(kconf)
         self.check_top_menu_not_too_long(kconf)
 
+    def get_modules(self, modules_file):
+        """
+        Get a list of modules and put them in a file that is parsed by
+        Kconfig
+
+        This is needed to complete Kconfig sanity tests.
+
```
ulfalizer: Super nit: Extra blank line.

scripts/check_compliance.py
```diff
@@ -252,7 +263,7 @@ def parse_kconfig(self):
         os.environ['GENERATED_DTS_BOARD_CONF'] = "dummy"
 
         # For multi repo support
-        get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
+        self.get_modules(os.path.join(tempfile.gettempdir(), "Kconfig.modules"))
```
ulfalizer: Could get rid of the `"Kconfig.modules"` argument and inline it into the function maybe.

scripts/check_compliance.py
```diff
@@ -223,6 +217,23 @@ def run(self):
...
+
+        """
+        try:
+            modules = sh.west("list", "--format={posixpath}", **sh_special_args)
+        except sh.CommandNotFound as e:
+            self.skip("Command not found: " + str(e))
+        except sh.ErrorReturnCode as e:
+            self.skip(e)
+
+        create_modules_file(modules, modules_file)
```
ulfalizer: Could inline this function here (and rename `get_modules()` to `create_modules_file()`). Doesn't seem to be any other users.

`get_*()` makes it sound like it returns a value to me, but I know this is old code.

nashif:
You know that sooner or later, not having west installed is not going to be an option. if you do not have west, you will have no modules file and Kconfig check will always fail..

marc-hb:
I understood there was a commitment to keep "some" (obviously: not Kconfig checks) basic build time features working without west, even in the long term? I think this commitment was recently renewed in other issue https://github.com/zephyrproject-rtos/zephyr/issues/14177 - which I didn't even expect to be addressed considering how minor it seemed.

Issue #20, opened 2019-02-23
Add checker for python code


nashif:
Check for python syntax using one of the many linters:
- flake8
- pylint
- ..

preferably this should work on diffs rather on complete files.

marc-hb:
BTW here's `pylint`'s most missing feature, especially in the context of focusing on new code and ignoring warnings on existing code:
```shell
usefulpylint() { pylint "$@" | sort -t ":" -k 2,2n ; }
usefulpylint -E scripts/sanitycheck
```

ulfalizer:
Only takes ~5 seconds or so run `pylint3` on all `.py` files in the tree, so for stuff that should always generate a failure, it's easy to do that instead.

marc-hb:
Duplicate of issue https://github.com/zephyrproject-rtos/zephyr/issues/10338 ?

PR #37, opened 2019-03-19
check_compliance.py: Run pylint on Python scripts

ulfalizer:
Run pylint on new/modified *.py files and fail if there are any
warnings.

Only enable a limited set of checks, focusing on logic issues and
relatively uncontroversial style stuff. We can tweak what checks are
enabled as we go.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,4 @@
 /scripts/check_compliance.py				@nashif
 /scripts/what_changed.py                                @nashif
+/scripts/pylintrc                                       @ulfalizer @mbolivar
 /CODEOWNERS						@nashif
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -671,6 +671,52 @@ class GitLint(ComplianceTest):
             self.add_failure(msg.decode("utf-8"))
 
 
+class PyLint(ComplianceTest):
+    """
+    Runs pylint on all .py files, with a limited set of checks enabled. The
+    configuration is in the pylintrc file.
+    """
+    _name = "pylint"
+    _doc = "https://www.pylint.org/"
+
+    def run(self):
+        self.prepare(ZEPHYR_BASE)
+
+        # Path to pylint configuration file
+        pylintrc = os.path.join(os.path.dirname(__file__), "pylintrc")
+
+        # List of .py files added/modified by the commit(s).
+        #
+        # Skip create_board_img.py to work around a crash in pylint 2.2.2:
+        # https://github.com/PyCQA/pylint/issues/2906
+        py_files = git(
+            "diff", "--name-only", "--diff-filter=d", self.commit_range,
+            "--", "*.py",
+            ":!boards/xtensa/intel_s1000_crb/support/create_board_img.py") \
+            .splitlines()
+
+        if not py_files:
+            return
+
+        try:
+            # Run pylint on added/modified .py files
+            process = subprocess.Popen(
+                ["pylint", "--rcfile=" + pylintrc] + py_files,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                cwd=ZEPHYR_BASE)
+        except FileNotFoundError:
+            self.error("pylint not found. Check that the directory it's in is "
+                       "listed in the PATH environment variable.")
+        except OSError as e:
+            self.error("Failed to run pylint: " + str(e))
+
+        stdout, stderr = process.communicate()
+        if process.returncode or stderr:
+            # Issues found, or a problem with pylint itself
+            self.add_failure(stdout.decode("utf-8") + stderr.decode("utf-8"))
+
+
 class License(ComplianceTest):
     """
     Checks for licenses in new files added by the Pull request
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -41,6 +41,10 @@ def git(*args):
     # Hack: Setting _cwd to the working dir seems pointless as of writing (it
     # should be the default), but keep it around in case it's working around
     # some issue
+
+    # pylint doesn't like how the sh library works.
+    #
+    # pylint: disable=unexpected-keyword-arg
     return sh.git(*args, _tty_out=False, _cwd=os.getcwd())
 
 
@@ -102,7 +106,6 @@ class ComplianceTest:
         Run testcase
         :return:
         """
-        pass
 
     def error(self, msg):
         """
--- a/scripts/pylintrc
+++ b/scripts/pylintrc
@@ -1,0 +1,246 @@
+# Copyright (c) 2019, Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+# pylint configuration for the PyLint check in check_compliance.py.
+#
+# To run pylint manually with this configuration from the Zephyr repo, do
+#
+#     pylint3 --rcfile=ci-tools/scripts/pylintrc <Python file>
+#
+# This command will check all scripts:
+#
+#     pylint3 --rcfile=ci-tools/scripts/pylintrc $(git ls-files '*.py')
+
+[MASTER]
+
+# Use multiple processes
+jobs=0
+
+# Do not pickle collected data for comparisons
+persistent=no
+
+
+[REPORTS]
+
+# Only show messages, not full report
+reports=no
+
+# Disable score
+score=no
+
+
+[MESSAGES CONTROL]
+
+# Only enable specific (hopefully) uncontroversial warnings. Use
+# 'pylint3 --list-msgs' to list messages and their IDs.
+#
+# These might be nice to check too, but currently trigger false positives:
+#
+#   no-member
+#   arguments-differ
+#   redefine-in-handler
+#   abstract-method
+#
+# These might be too controversial:
+#
+#   no-else-return
+#   consider-using-get
+#   redefined-builtin
+#
+# These tell you to use logger.warning("foo %d bar", 3) instead of e.g.
+# logger.warning("foo {} bar".format(3)), but it's not a clear win in all
+# cases. f-strings would be nicer too, and it's easier to convert from format()
+# to those.
+#
+#   logging-not-lazy
+#   logging-format-interpolation
+#   logging-fstring-interpolation
+
+disable=all
+# Identifiers are in the same order as in 'pylint3 --list-msgs'
+enable=
+    empty-docstring,
+    unneeded-not,
+    singleton-comparison,
+    misplaced-comparison-constant,
+    unidiomatic-typecheck,
+    consider-using-enumerate,
+    consider-iterating-dictionary,
+    bad-classmethod-argument,
+    bad-mcs-method-argument,
+    bad-mcs-classmethod-argument,
+    single-string-used-for-slots,
+    trailing-newlines,
+    trailing-whitespace,
+    missing-final-newline,
+    superfluous-parens,
+    mixed-line-endings,
+    unexpected-line-ending-format,
+    invalid-characters-in-docstring,
+    useless-import-alias,
+    len-as-condition,
+    syntax-error,
+    init-is-generator,
+    return-in-init,
+    function-redefined,
+    not-in-loop,
+    return-outside-function,
+    yield-outside-function,
+    nonexistent-operator,
+    duplicate-argument-name,
+    abstract-class-instantiated,
+    bad-reversed-sequence,
+    too-many-star-expressions,
+    invalid-star-assignment-target,
+    star-needs-assignment-target,
+    nonlocal-and-global,
+    continue-in-finally,
+    nonlocal-without-binding,
+    misplaced-format-function,
+    method-hidden,
+    access-member-before-definition,
+    no-method-argument,
+    no-self-argument,
+    invalid-slots-object,
+    assigning-non-slot,
+    invalid-slots,
+    inherit-non-class,
+    inconsistent-mro,
+    duplicate-bases,
+    non-iterator-returned,
+    unexpected-special-method-signature,
+    invalid-length-returned,
+    relative-beyond-top-level,
+    used-before-assignment,
+    undefined-variable,
+    undefined-all-variable,
+    invalid-all-object,
+    no-name-in-module,
+    unpacking-non-sequence,
+    bad-except-order,
+    raising-bad-type,
+    bad-exception-context,
+    misplaced-bare-raise,
+    raising-non-exception,
+    notimplemented-raised,
+    catching-non-exception,
+    bad-super-call,
+    not-callable,
+    assignment-from-no-return,
+    no-value-for-parameter,
+    too-many-function-args,
+    unexpected-keyword-arg,
+    redundant-keyword-arg,
+    missing-kwoa,
+    invalid-sequence-index,
+    invalid-slice-index,
+    assignment-from-none,
+    not-context-manager,
+    invalid-unary-operand-type,
+    unsupported-binary-operation,
+    repeated-keyword,
+    not-an-iterable,
+    not-a-mapping,
+    unsupported-membership-test,
+    unsubscriptable-object,
+    unsupported-assignment-operation,
+    unsupported-delete-operation,
+    invalid-metaclass,
+    unhashable-dict-key,
+    logging-unsupported-format,
+    logging-format-truncated,
+    logging-too-many-args,
+    logging-too-few-args,
+    bad-format-character,
+    truncated-format-string,
+    mixed-format-string,
+    format-needs-mapping,
+    missing-format-string-key,
+    too-many-format-args,
+    too-few-format-args,
+    bad-string-format-type,
+    bad-str-strip-call,
+    invalid-envvar-value,
+    yield-inside-async-function,
+    not-async-context-manager,
+    useless-suppression,
+    deprecated-pragma,
+    use-symbolic-message-instead,
+    literal-comparison,
+    comparison-with-itself,
+    no-self-use,
+    no-classmethod-decorator,
+    no-staticmethod-decorator,
+    cyclic-import,
+    duplicate-code,
+    consider-merging-isinstance,
+    simplifiable-if-statement,
+    redefined-argument-from-local,
+    trailing-comma-tuple,
+    stop-iteration-return,
+    useless-return,
+    consider-swap-variables,
+    consider-using-join,
+    consider-using-in,
+    chained-comparison,
+    consider-using-dict-comprehension,
+    consider-using-set-comprehension,
+    simplifiable-if-expression,
+    unreachable,
+    pointless-statement,
+    pointless-string-statement,
+    expression-not-assigned,
+    unnecessary-pass,
+    unnecessary-lambda,
+    duplicate-key,
+    assign-to-new-keyword,
+    useless-else-on-loop,
+    confusing-with-statement,
+    using-constant-test,
+    comparison-with-callable,
+    lost-exception,
+    assert-on-tuple,
+    bad-staticmethod-argument,
+    super-init-not-called,
+    non-parent-init-called,
+    useless-super-delegation,
+    unnecessary-semicolon,
+    bad-indentation,
+    mixed-indentation,
+    deprecated-module,
+    reimported,
+    import-self,
+    misplaced-future,
+    global-variable-not-assigned,
+    unused-import,
+    unused-variable,
+    undefined-loop-variable,
+    unbalanced-tuple-unpacking,
+    possibly-unused-variable,
+    self-cls-assignment,
+    bare-except,
+    duplicate-except,
+    try-except-raise,
+    binary-op-exception,
+    raising-format-tuple,
+    wrong-exception-operation,
+    keyword-arg-before-vararg,
+    bad-format-string-key,
+    unused-format-string-key,
+    bad-format-string,
+    unused-format-string-argument,
+    format-combined-specification,
+    missing-format-attribute,
+    invalid-format-index,
+    anomalous-backslash-in-string,
+    anomalous-unicode-escape-in-string,
+    bad-open-mode,
+    redundant-unittest-assert,
+    deprecated-method,
+    bad-thread-instantiation,
+    shallow-copy-environ,
+    invalid-envvar-default,
+    deprecated-string-function,
+    deprecated-str-translate-call,
+    deprecated-itertools-function,
+    deprecated-types-field,


ulfalizer:
Work in progress, but feel free to complain about stuff.

ulfalizer:
Wonder why CI gets stuck...

@marc-hb 
Did you figure out how to see the logs?

nashif:
> Wonder why CI gets stuck...

```
${CITOOLS}/scripts/merge_junit.py first_pass.xml second_pass.xml > compliance.xml
Traceback (most recent call last):
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/merge_junit.py", line 63, in <module>
    main()
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/merge_junit.py", line 26, in main
    merge_results(args[:])
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/merge_junit.py", line 37, in merge_results
    tree = ET.parse(file_name)
  File "/usr/lib/python3.5/xml/etree/ElementTree.py", line 1184, in parse
    tree.parse(source, parser)
  File "/usr/lib/python3.5/xml/etree/ElementTree.py", line 585, in parse
    source = open(source, "rb")
FileNotFoundError: [Errno 2] No such file or directory: 'first_pass.xml'
```

nashif:
```
Traceback (most recent call last):
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/check_compliance.py", line 897, in <module>
    main()
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/check_compliance.py", line 853, in main
    test.run()
  File "/var/lib/shippable/build/IN/ci_tools_repo/gitRepo/scripts/check_compliance.py", line 451, in run
    cwd=self.zephyr_base)
  File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'pylint3'
```

nashif:
you probably want to call pylint instead of pylint3?

ulfalizer:
@nashif 
`pylint` is Python 2 only, at least on my Ubuntu system. I get errors when I run it on code that uses Python 3-only features.

Maybe pylint3 needs to be installed on the CI machines.

marc-hb:
> pylint is Python 2 only, at least on my Ubuntu system. I get errors when I run it on code that uses Python 3-only features.

Just like pip, python and other python tools, on some Linux distributions pylint is 2 and on others it's 3.

You need something like: "pylint3 if pylint3 else pylint", I implemented this solution on some other (Zephyr) places and I haven't seen any better way.

ulfalizer:
@marc-hb @nashif 
Do you know which version `pylint` is on our CI systems?

If it's the Python 2 version, then that'll need to be fixed at least.

marc-hb:
Interesting but I'd much rather have something that runs MOST `pylint` and `pep8` checks (not just "unused" checks) on just *changed* files and highlights changed lines on a per commit basis, think large series.

The purpose of `check_compliance.py` is to provide fast feedback on submitted commits, scanning and cleaning up the entire source code is also useful but it doesn't belong to this script.

Links: issue #20

ulfalizer:
@marc-hb 
Could have both. Getting rid of all unused imports and variables everywhere forever is a great first step. :)

marc-hb:
> Could have both.

Yes, one in this script and the other not.

BTW attempts and plans to "keep the entire codebase always clean at all times" I've observed in past projects have always failed miserably when "making contact with the enemy" and always end up in a massive amount of noise in the git log making git blame, cherry-pick etc. unusable. For instance what happens when some new pylint version fixes a bug and start reporting many more errors?

ulfalizer:
@marc-hb 
Hasn't been a problem in practice for the Kconfig checks at least. If it turns into a pain for pylint, we could always do something about it then.

The thing I like about always checking everything is that it's super simple, and keeps people honest (no sneaking past it :).

Changed `pylint3` to `pylint`. Let's see what happens now.

New batch fix: https://github.com/zephyrproject-rtos/zephyr/pull/14754

ulfalizer:
@nashif 
What's the log now? It ran at least, but I need to see how it failed.

Would be nice if the logs were linked from here automatically.

ulfalizer:
I like that `check_compliance.py` is nitting itself now. :D

marc-hb:
> The thing I like about always checking everything is that it's super simple, and keeps people honest (no sneaking past it :).

It's OK for simple stuff like undefined references and unused things, not so much when you start looking at actual code *style* which is the primary focus of pylint and pep8 and never 100% "reliable"

ulfalizer:
> > The thing I like about always checking everything is that it's super simple, and keeps people honest (no sneaking past it :).
> 
> It's OK for simple stuff like undefined references and unused things, not so much when you start looking at actual code _style_ which is the primary focus of pylint and pep8 and never 100% "reliable"

Yeah, I want to focus on the uncontroversial stuff first, that can always safely be flagged. As long as the code doesn't look too horrible, I think that's more important than style nits.

My previous experience with overly-strict style checks is that they waste more time than they help, and distract from things that are much more important when it comes to code clarity.

Could maybe get some suggestion system in there though. For uncontroversial stuff, I think it's better to just fail.

marc-hb:
Illustration on Slack yesterday:
> andrewboie [11:50 AM]
> @ galak we have some commits by @ulfalizer which clean up syntax/style for a bunch of Python scripts. Do we want to merge those for 1.14 or wait to 1.15? my thought was that if they go into 1.14, this will greatly ease backporting any other fixes to those scripts to the LTS branch

marc-hb:
> My previous experience with overly-strict style checks is that they waste more time than they help, and distract from things that are much more important when it comes to code clarity.

"Overly strict" can refer to two totally different things: 1. "pixel perfection" = the rules themselves, 2. the "appeal process" to deal with exceptions.

"Soft" codestyle checks that make sense only 90 +-5% of the times are still very useful  as long as:
- their new versions and/or configurations not regularly carpet-bombed on the whole codebase as "offered" by people who don't know or don't care  about version control.
- they can easily be ignored when needed

https://www.kernel.org/doc/html/v4.12/process/5.Posting.html#sending-the-patch
> Are you sure your patch is free of silly mistakes? You should always run patches through scripts/checkpatch.pl and address the complaints it comes up with. Please bear in mind that checkpatch.pl, while being the embodiment of a fair amount of thought about what kernel patches should look like, is not smarter than you. If fixing a checkpatch.pl complaint would make the code worse, don’t do it.

ulfalizer:
@marc-hb 
Could get some soft checks in there later maybe. For now, failing for uncontroversial stuff is a good start. Don't have to do all of it at once, and it's nice to plug stuff as early as possible, to make them someone else's problem. ;)

marc-hb:
Most of Pylint/pep8 are soft checks. I know because I've (tried to) run pylint on existing scripts in this repo and by default the number of warnings was so high it was unusable.
So the class that enforces across the entire codebase the small number of Pylint _hard checks_ should not be called "Pylint". Maybe that's the point I've trying to make from the start?
`class PylintHardChecks`?

ulfalizer:
Not against renaming it later, when/if we add more checks.

ulfalizer:
@nashif 
@marc-hb 
Should be ready now, once https://github.com/zephyrproject-rtos/zephyr/pull/15962 is in.

ulfalizer:
@nashif 
Where do you get the log for "error during verification" stuff?

scripts/check_compliance.py
```diff
@@ -588,6 +588,118 @@ def run(self):
...
+        """
+        Returns a list with the paths of all .py files, relative to the Zephyr
+        root.
+
+        Uses 'git ls-files', which is fast and avoids picking up any files that
+        are not checked in.
+        """
+        cmd = "git ls-files -z *.py"  # -z = null-separate entries
+
+        process = subprocess.Popen(cmd.split(),
```
marc-hb: Anything wrong with subprocess.check_output()?

scripts/check_compliance.py
```diff
@@ -588,6 +588,118 @@ def run(self):
...
+
+        stdout, stderr = process.communicate()
+        if process.returncode or stderr:
+            # Issues found, or a problem with pylint itself
+            self.add_failure(stdout.decode("utf-8") + stderr.decode("utf-8"))
+
+
+    def py_files(self):
+        """
+        Returns a list with the paths of all .py files, relative to the Zephyr
```
marc-hb: Running on unmodified files is a waste of time, the purpose of this script is to check incremental commits. Other check(s) use `git diff --name-only`, it should be possible to refactor and re-use.

scripts/check_compliance.py
```diff
@@ -588,6 +588,118 @@ def run(self):
...
+    _doc = "https://www.pylint.org/"
+
+    def run(self):
+        self.prepare(ZEPHYR_BASE)
+
+        # The commented-out checks below might be useful to enable, but
+        # currently trigger false positives
+        cmd = ("pylint --jobs=0 --score=n --persistent=n --disable=all "
+               "--enable "
+               "C0303,"  # Trailing whitspace
```
marc-hb: Can you move this configuration to some `--rcfile=pylint-requirements`?

This may help get rid of the numerical IDs.

scripts/check_compliance.py
```diff
@@ -588,6 +588,118 @@ def run(self):
             self.add_failure(msg.decode("utf-8"))
 
 
+class PyLint(ComplianceTest):
```
marc-hb: How about `class PylintRequirements( )` to make clearer that the vast majority of Pylint checks are neither required nor even run?

scripts/check_compliance.py
ulfalizer: Gonna make a commit later to move this repo back to `scripts/ci` in the main Zephyr repo.

scripts/check_compliance.py
ulfalizer: I wouldn't have any idea what `PylintRequirements` means personally. It runs `pylint`. :)

scripts/check_compliance.py
ulfalizer: It's very simple, rock solid, and fast enough.

scripts/check_compliance.py
marc-hb: Also please test and make sure a missing pylint doesn't stop other, unrelated checks.

ulfalizer: Tried a rewrite now.

One thing that gets less clean compared to `Popen()` is failing if there's anything on `stderr`, even if the return status is 0, without redirecting `stderr`.

No value in using `check_output()` for its own sake, IMO, if `Popen()` gives you a little bit if added flexibility that you make use of. The `Popen()` interface isn't all that bad either.

scripts/check_compliance.py
ulfalizer: Ohh... you mean as an argument to `pylint`.

IMO, it's nicer to have directly in the code. Cuts down on the number of files and makes the list easy to discover.

One downside is that people might mess up the formatting, but maybe the simplicity outweighs that.

ulfalizer: Could maybe make it a triple-quoted string or the like to avoid that though, and `','.join(pylint_checks)` to automatically insert the `,`s. Maybe just a string is good enough though.

scripts/check_compliance.py
ulfalizer: Now checks for troubles starting `pylint`.

Realized even `subprocess.check_output()` raises `OSError` if the executable is missing btw, so there's probably other places that ought to catch it.

ulfalizer:
No idea what's going on with the test atm. Is there something nicer than randomly commenting out and re-submitting to troubleshoot it with?

scripts/check_compliance.py
```diff
@@ -588,6 +588,122 @@ def run(self):
...
+    """
+    Runs pylint on all .py files, with a limited set of checks enabled
+    """
+    _name = "pylint"
```
marc-hb: The other checks are in CamelCase. I don't like CamelCase in CLIs (-m pylint) but I like losing time because of the inconsistency even less.

scripts/check_compliance.py
marc-hb: > It's very simple, rock solid

The `git diff --name-only` command has already been "in production" for months so it's rock solid too and barely longer. It also reflects the general purpose of this script as documented in its `--commits` argument help which this check silently drops on the floor.

> and fast enough

check_compliance.py -m pylint lasts 7 seconds and uses 52 seconds of CPU time on my workstation for a zero python change commit so: no, it's not "fast enough". @nashif may also have an opinion about the Shippable bill.

> My modus operandi is to clean up the entire codebase first

Great, once that's done you don't need to have every Shippable instance repeatedly and pointlessly scan the same, unchanged python code over and over.  If the config file is extracted from the code then anyone who wants to scan the whole codebase again [*] can at anytime, they don't even need to run this script.

[*] for instance to evolve the config incrementally without requiring a "big bang" fix.

ulfalizer: Alright, I'll change it. It usually means you end up with failures getting in through merges and CI overrides and stuff over time, so I think it's a bad idea though.

scripts/check_compliance.py
ulfalizer: It should be the same thing. The only thing that differs is in which file you edit the options.

There's a general trend of looking at anything that seems too simple and direct with suspicion in modern software dev. Stupidly simple is freaking awesome. :ok_hand:

scripts/check_compliance.py
```diff
@@ -588,6 +588,122 @@ def run(self):
...
+               "E0101,"  # Explicit return in __init__
+               "E0102,"  # Function/method/class defined twice
+               "E0107,"  # Use of non-existent operators
+               "E0202,"  # Attribute named the same as method
+               "E0203,"  # Access to member before definition
+               "E0211,"  # Method has no argument
+               "E0213,"  # Method should have self as first argument
+               "E0221,"  # Interface resolved to x is not a class
+               "E0222,"  # Missing method from interface
+               #"E0601,"  # Variable referenced before assignment
```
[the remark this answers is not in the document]
ulfalizer: There's a comment at the top that explains why they're commented out, just in case you missed it.

I could move the list of disabled checks somewhere else though if you'd like.

scripts/check_compliance.py
```diff
@@ -588,6 +588,122 @@ def run(self):
...
+            self.error("Workaround in check_compliance.py no longer needed? "
+                       "Please check (grep for this message).")
+
+        try:
+            process = subprocess.Popen(cmd.split() + py_files,
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE,
+                                       cwd=ZEPHYR_BASE)
+        except OSError as e:
+            # pylint missing, most likely
```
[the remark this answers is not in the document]
ulfalizer: Fixed by using a different message for `FileNotFoundError`.

scripts/check_compliance.py
ulfalizer: There's `_name = "checkpatch"` too. Could always change them all to lowercase instead.

ulfalizer:
I'll do a separate PR with a helper function for getting the output of a command and failing the test and printing a `PATH` hint, etc., if it can't be found.

Could remove the `sh` library dependency at the same time. It's probably overkill just for running `git`.

Thinking something like this:

```python
output = self.run("cmd", "arg", "arg", context="trying to do stuff")
```

scripts/check_compliance.py
marc-hb: The nightly builds are the best place to catch these, so they don't confuse and bother submitters with failures totally unrelated to their submission. Assuming a separate config file it should be as simple as adding literally one line in the nightly builds.

ulfalizer: Now runs `pylint` on just new/modified Python scripts.

scripts/check_compliance.py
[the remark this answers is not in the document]
ulfalizer: Also, I'm not 100% on having this be a whitelist instead of a blacklist (that one might become shorter too, though I haven't checked). With a blacklist, you might have to tweak it depending on `pylint` version though.

It's easy to tweak later at least.

scripts/check_compliance.py
marc-hb: > The only thing that differs is in which file you edit the options.

If configuration and code are in the same file then you cannot simply "git checkout" only one of them and test. You cannot git log the config file to look at just the configuration history. You cannot run pylint without this script. You cannot easily compare your configuration to someone else or pass it / share it with some other tool. Etc., etc.

I really don't see what's complicated about separating configuration from the code, it's a standard (.checkpatch.conf, .gitlint, CODEOWNERS,...) and massively successful approach and I don't remember anyone complaining about it before.

scripts/check_compliance.py
ulfalizer: Is there some way to add a GitHub message without also failing the build btw?

If we add some thing to give the user the complete `pylint` output later, then maybe it could be implemented in the same class.

Thinking something like

```python
self.message("Here is the complete pylint output. Check for anything that "
             "should be fixed.\n\n" + full_pylint_output)
```

scripts/check_compliance.py
ulfalizer: @marc-hb 
Refresh ;)

ulfalizer: Or do you still want it in the nightly builds?

The Python files are known to be warning-free now, so it shouldn't be possible to get false positives, unless something slips through through a forced merge or the like.

scripts/check_compliance.py
marc-hb: Fears of breaking backward-compatibility and/or CI aside, I'd love everything to become lowercase. Not in this commit anyway.

scripts/check_compliance.py
ulfalizer: Could argue that \<anything> should be moved into a separate file that way though ("if X and Y are in the same file you cannot simply...").

Do you think it's common to need the whitelist from commit X but not the rest of the test (which is still simple to do)? Extra files by themselves are a liability, so nice to skip them otherwise.

> You cannot run pylint without this script

Why not run the test itself if you want to do what the test does? (Genuine question... might be something I've missed.)

> I really don't see what's complicated about separating configuration from the code, it's a standard (.checkpatch.conf, .gitlint, CODEOWNERS,...) and massively successful approach and I don't remember anyone complaining about it before.

Here you already know what the configuration is. It's just moving it into a separate file.

Whether it's a list of parameters to some external command or some other data associated with the test makes no difference to me.

scripts/check_compliance.py
ulfalizer: Could I keep it lowercase and lowercase the other ones later? Doesn't seem to be consistent at the moment at least.

ulfalizer:
@marc-hb 
Moved the configuration into a separate configuration file.

scripts/check_compliance.py
```diff
@@ -668,6 +668,51 @@ def run(self):
...
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                cwd=ZEPHYR_BASE)
+        except FileNotFoundError:
+            self.error("pylint not found. Check that the directory it's in is "
+                       "listed in the PATH environment variable.")
+        except OSError as e:
+            self.error("Failed to run pylint: " + str(e))
+
+        stdout, stderr = process.communicate()
```
marc-hb: The more recent (Python 3.5) `subprocess.run()` saves you the `.communicate()` and manual utf-8 decoding.

scripts/check_compliance.py
```diff
@@ -668,6 +668,51 @@ def run(self):
...
+
+        # Path to pylint configuration file
+        pylintrc = os.path.join(os.path.dirname(__file__), "pylintrc")
+
+        # List of .py files added/modified by the commit(s).
+        #
+        # Skip create_board_img.py to work around a crash in pylint 2.2.2:
+        # https://github.com/PyCQA/pylint/issues/2906
+        py_files = git(
+            "diff", "--name-only", self.commit_range, "--", "*.py",
```
marc-hb: Does this not fail to run when renaming or deleting a file? See ``--diff-filter` in other checks.

ulfalizer:
> Could you elaborate a tiny bit more about the initial config in the commit message or in the config file itself? For instance: did you start from some other, existing project or some PEP recommentation(s)? How did it evolve from that? Roughly how many warnings does this initial config file report if used on the entire and current master branch? Etc.

I enable everything that seems noncontroversial, and fix all the reported stuff (there's lots of PRs in the Zephyr repo).

`pylint` is more of a linter than a style checker, so not sure there are any PEPs or recommendations or anything like that to refer to.

scripts/check_compliance.py
ulfalizer: The `encoding` parameter to `subprocess.run()` was added in Python 3.6, and that's too recent, so won't get rid of the decoding I think. Think we might still support Python 3.4 too.

I like the high-level interface personally. Once you've gotten it right, it's easy to tweak.

scripts/check_compliance.py
ulfalizer: Oops... fixed. Went with `--diff-filter=d` (everything but deleted). Looked like the rest would be safe to check, and maybe it could find stuff we haven't thought of.

scripts/check_compliance.py
mbolivar: > Think we might still support Python 3.4 too.

Correct: https://github.com/zephyrproject-rtos/zephyr/blob/master/cmake/python.cmake#L16

I'd like to drop 3.4 support, as it's EOL in upstream Python since March [*], and I've seen a fair bit of useful stuff added in 3.5 that requires workarounds or less elegant methods in 3.4. I haven't been able to figure out if we've got any older RHEL or Ubuntu 14.04 users out there that would complain, though.

[*] https://www.python.org/downloads/release/python-3410/

scripts/check_compliance.py
marc-hb: (Just for the record)

For renames that have been detected (depending on the curent similarity threshold), it seems `--name-only` prints only the new name. Interesting!

CODEOWNERS
```diff
@@ -1,3 +1,4 @@
 /scripts/check_compliance.py				@nashif
 /scripts/what_changed.py                                @nashif
+/scripts/pylintrc                                       @ulfalizer
```
marc-hb: Any other volunteer? Style issues can get easily get controversial...

scripts/check_compliance.py
```diff
@@ -41,6 +41,10 @@ def git(*args):
     # Hack: Setting _cwd to the working dir seems pointless as of writing (it
     # should be the default), but keep it around in case it's working around
     # some issue
+
+    # pylint doesn't like how the sh library works.
+    #
+    # pylint: disable=unexpected-keyword-arg
```
marc-hb: BTW "sh" is the only module that doesn't work on Windows AFAIR. Off-topic sorry.

scripts/check_compliance.py
```diff
@@ -668,6 +671,52 @@ def run(self):
...
+    configuration is in the pylintrc file.
+    """
+    _name = "pylint"
+    _doc = "https://www.pylint.org/"
+
+    def run(self):
+        self.prepare(ZEPHYR_BASE)
+
+        # Path to pylint configuration file
+        pylintrc = os.path.join(os.path.dirname(__file__), "pylintrc")
```
marc-hb: How about ZEPHYR_BASE instead of __file__?

ulfalizer: `ZEPHYR_BASE` points to the Zephyr root rather than the `scripts` directory in the `ci-tools` repository, so don't think it will work.

Think you're allowed to have the `ci-tools` repository checked out anywhere relative to `ZEPHYR_BASE`.

CODEOWNERS
ulfalizer: Could put you in if you want to.

ulfalizer: Can run `pylint --list-msgs` to see a list of messages btw. Most of them are enabled.

Types of stuff I skipped:

- Wrong hanging indent, because it'd require reformatting a ton of code, and it's too naggy. It 
  also flags deliberate alignment.

- Whitespace errors like `x = [ 1, 2, 3 ]` (instead of `x = [1, 2, 3]`). Too naggy, flags deliberate alignment. (I did a whitespace cleanup pass in a commit to fix all stuff like in the example though.)

- Stuff like avoiding `global` and other language features altogether. Too dogmatic.

- Naming stuff like never using super short variable names, and uppercasing global variables. Too dogmatic and naggy. Short names are fine if they're clear from context.

- Preferring

   ```python
   if foo:
       return bar
   return baz
   ```

   over

   ```python
   if foo:
       return bar
   else:
       return baz
   ```

   I like that one, but it might not always be a win, and it's too subjective.

scripts/check_compliance.py
marc-hb: Doh! Yes of course. Sorry for the noise.

CODEOWNERS
ulfalizer: @marc-hb 
Added @mbolivar to CODEOWNERS.

PR #42, opened 2019-04-02
check_compliance: Use zephyr_module.py to generate Kconfig.modules

carlescufi:
Instead of duplicating the code already present in Zephyr, reuse the
existing scripts/zephyr_module.py to generate the Kconfig.modules file
required by Kconfig.

Signed-off-by: Carles Cufi <carles.cufi@nordicsemi.no>


marc-hb:
Not so good: PR #47

PR #43, opened 2019-04-07
license: check for CMakeLists.txt and Kconfig files

nashif:
Check license for cmake and kconfig files.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -578,33 +578,51 @@ class License(ComplianceTest):
 
         report = ""
 
-        whitelist_extensions =  ['.yaml', '.html']
-        whitelist_languages = ['CMake', 'HTML']
+        never_check_ext =  ['.yaml', '.html', '.rst', '.conf', '.cfg']
+        never_check_langs = ['HTML']
+        check_langs = ['CMake']
         with open('scancode.json', 'r') as json_fp:
             scancode_results = json.load(json_fp)
             for file in scancode_results['files']:
                 if file['type'] == 'directory':
                     continue
 
-                original_fp = str(file['path']).replace('scancode-files/', '')
+                orig_path = str(file['path']).replace('scancode-files/', '')
                 licenses = file['licenses']
-                if (file['is_script'] or file['is_source']) and (file['programming_language'] not in whitelist_languages) and (file['extension'] not in whitelist_extensions):
-                    if not file['licenses']:
-                        report += ("* {} missing license.\n".format(original_fp))
+                file_type = file.get("file_type")
+                kconfig = "Kconfig" in orig_path and file_type in ['ASCII text']
+                check = False
+
+                if file.get("extension") in never_check_ext:
+                    check = False
+                elif file.get("programming_language") in never_check_langs:
+                    check = False
+                elif kconfig:
+                    check = True
+                elif file.get("programming_language") in check_langs:
+                    check = True
+                elif file.get("is_script"):
+                    check = True
+                elif file.get("is_source"):
+                    check = True
+
+                if check:
+                    if not licenses:
+                        report += ("* {} missing license.\n".format(orig_path))
                     else:
                         for lic in licenses:
                             if lic['key'] != "apache-2.0":
                                 report += ("* {} is not apache-2.0 licensed: {}\n".format(
-                                    original_fp, lic['key']))
+                                    orig_path, lic['key']))
                             if lic['category'] != 'Permissive':
                                 report += ("* {} has non-permissive license: {}\n".format(
-                                    original_fp, lic['key']))
+                                    orig_path, lic['key']))
                             if lic['key'] == 'unknown-spdx':
                                 report += ("* {} has unknown SPDX: {}\n".format(
-                                    original_fp, lic['key']))
+                                    orig_path, lic['key']))
 
                     if not file['copyrights']:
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -588,7 +589,16 @@ def run(self):
...
                 licenses = file['licenses']
-                if (file['is_script'] or file['is_source']) and (file['programming_language'] not in whitelist_languages) and (file['extension'] not in whitelist_extensions):
+                is_script = file.get("is_script")
+                is_source = file.get("is_source")
+                is_media = file.get("is_media")
+                pl = file.get("programming_language", None)
+                extension = file.get("extension")
+                file_type = file.get("file_type")
+                kconfig = "Kconfig" in original_fp and file_type in ['ASCII text']
+                valid_type = is_script or is_source or pl in blacklist_languages or not is_media or kconfig
```
ulfalizer: This will be true for any file with `not file.get("is_media")`. Is that deliberate?

Could maybe use this style instead of creating lots of variables that are only used in a single place:

```python
checked_file = (file.get('is_script') or
                file.get('is_source') or
                ...)
```

Makes it clear where the value is used, and that it's only used once.

scripts/check_compliance.py
```diff
@@ -588,7 +589,16 @@ def run(self):
 
                 original_fp = str(file['path']).replace('scancode-files/', '')
                 licenses = file['licenses']
-                if (file['is_script'] or file['is_source']) and (file['programming_language'] not in whitelist_languages) and (file['extension'] not in whitelist_extensions):
+                is_script = file.get("is_script")
+                is_source = file.get("is_source")
+                is_media = file.get("is_media")
+                pl = file.get("programming_language", None)
```
ulfalizer: `lang` might be clearer than `pl`.

scripts/check_compliance.py
```diff
@@ -588,7 +589,16 @@ def run(self):
...
+                extension = file.get("extension")
+                file_type = file.get("file_type")
+                kconfig = "Kconfig" in original_fp and file_type in ['ASCII text']
```
ulfalizer: Had to read back to understand that "fp" stands for "file path". Could call it `original_file_path` instead maybe, or just `original_path`.

scripts/check_compliance.py
nashif: > This will be true for any file with `not file.get("is_media")`. Is that deliberate?

yes, I do not want to report anything related to media files.

ulfalizer: But you want to report everything for everything that's not a media file? That's what the logic is now.

If scripts aren't media files, then those would automatically get included by the `or not is_media`, for example.

scripts/check_compliance.py
```diff
@@ -579,32 +579,48 @@ def run(self):
...
                 if file['type'] == 'directory':
                     continue
 
-                original_fp = str(file['path']).replace('scancode-files/', '')
+                orig_path = str(file['path']).replace('scancode-files/', '')
                 licenses = file['licenses']
-                if (file['is_script'] or file['is_source']) and (file['programming_language'] not in whitelist_languages) and (file['extension'] not in whitelist_extensions):
+                lang = file.get("programming_language", None)
+                extension = file.get("extension")
+                file_type = file.get("file_type")
```
ulfalizer: Could get rid of `extension` and `file_type` too. Single-use, and the value is as easy to understand as the name.

scripts/check_compliance.py
nashif: ah, you are right.

scripts/check_compliance.py
```diff
@@ -579,32 +579,47 @@ def run(self):
...
+                kconfig = "Kconfig" in orig_path and file_type in ['ASCII text']
+                checked_file = (
+                        file.get("is_script") or
+                        file.get("is_source") or
```
ulfalizer: If `is_script` or `is_source` is True for a file, then `blacklist_languages` and `whitelist_languages` won't matter. Is that deliberate?

scripts/check_compliance.py
```diff
@@ -578,33 +578,51 @@ def run(self):
 
         report = ""
 
-        whitelist_extensions =  ['.yaml', '.html']
-        whitelist_languages = ['CMake', 'HTML']
+        whitelist_extensions =  ['.yaml', '.html', '.rst', '.conf', '.cfg']
+        whitelist_languages = ['HTML']
+        blacklist_languages = ['CMake']
```
ulfalizer: Bit tricky (for me at least) to understand the intent here.

What does it mean if a language is in neither the whitelist or the blacklist? What if it's in both? Maybe a comment could be added to explain it.

Is there a more explicit name than "whitelist"/"blacklist" that could be used (like `always_checked_langs` and `never_checked_langs` or the like)? I keep getting confused about whether "whitelist" means to "whitelist it for the test" or "whitelist it so that the test doesn't check it". :)

scripts/check_compliance.py
```diff
@@ -578,33 +578,51 @@ def run(self):
...
+                elif file.get("programming_language") in never_check_langs:
+                    check = False
+                elif kconfig:
+                    check = True
+                elif file.get("programming_language") in check_langs:
+                    check = True
+                elif file.get("is_script"):
+                    check = True
+                elif file.get("is_source"):
+                    check = True
```
ulfalizer: Some kind of helper function might be nice here later.

PR #46, opened 2019-04-11
check_compliance.py: add comment explaining why Kconfig prints twice

marc-hb:
Also explain why we can't just "warn_to_stderr=False" our way out of it.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -257,6 +257,10 @@ class KconfigCheck(ComplianceTest):
         os.environ["KCONFIG_WARN_UNDEF"] = "y"
 
         try:
+            # Note this will both print warnings to stderr _and_ return
+            # them: so some warnings might get printed
+            # twice. "warn_to_stderr=False" could unfortunately cause
+            # some (other) warnings to never be printed.
             return kconfiglib.Kconfig()
         except kconfiglib.KconfigError as e:
             self.add_failure(str(e))


scripts/check_compliance.py
```diff
@@ -257,6 +257,10 @@ def parse_kconfig(self):
         os.environ["KCONFIG_WARN_UNDEF"] = "y"
 
         try:
+            # Note this will both print warnings to stderr _and_ return
+            # them: so some warnings might get printed
+            # twice. "warn_to_stderr=False" could unfortunately cause
+            # some (other) warnings to be printed never.
```
ulfalizer: Typo:

*to never be printed

marc-hb: This is meant to be a stylistic effect, heard that in some movie(s) or song(s)... @dbkinder to the rescue?

marc-hb: "printed once, printed twice, printed never"

ulfalizer: Yeah, sounds fine there at least.

dbkinder: sigh.  "never" should not be at the end of a clause or sentence, and should be before the main verb of the sentence:  "He's never invited to dinner" and **not** "He's invited to dinner never".  So @ulfalizer is right:  ``cause some warnings to never be printed.``  (The "don't use split infinitives" rule isn't strictly followed these days: putting an adverb between ``to`` and the verb.)

ulfalizer: Maybe we're to blame for the split infinitives, going on [this video](https://www.youtube.com/watch?v=CDAU3TpunwM&t=14m10s). ;)

marc-hb: Try to be a poet and get free English classes - what's not to like! Thanks @dbkinder.

(Now I hope no one looks at the commit message. github doesn't support reviewing and tracking commit messages; that's a good start)

dbkinder: Love the video.  Thanks for sharing.

PR #52, opened 2019-04-23
check_compliance.py: Add CONFIG_CRC to whitelist

bwitherspoon:
This fixes Kconfig issues for PR #15302. CONFIG_CRC is defined in the TI CC13x2
/ CC26x2 SDK.


bwitherspoon:
> Would a `CFG_` prefix work instead btw? Just checking.

Yes it would. I was hesitant to modify any sources in the SDK, but it turns out the offending CONFIG_CRC is in comments. I am going to just change the prefix and close this PR.

bwitherspoon:
Reopening. After review of zephyrproject-rtos/zephyr#15302, it was requested for check_compliance.py to be updated to resolve the Kconfig issues instead of modifying the SDK source.

PR #54, opened 2019-04-26
Codeowners: replace glob with pathlib. Many other fixes.

marc-hb:
The initial purpose was to replace glob.glob(**, recursive=True) with
pathlib.Path.glob(**) because the former follows symbolic links to
directories in recursive mode (including "infinite" loops), whereas the
latter (and more modern) doesn't. While trying to stick to this change,
testing found a significant number of other issues and addressing all of
them unfortunately turned into this significant rewrite. The entire
Codeowners check is too small to isolate every change in separate
commit, sorry for the resulting "bulk" review work load (~100 lines)

Here's a simplified and extreme reproduction:
- run a fairly extensive sanitycheck run. This creates 100,000s of files
  in sanity-out (thank you cmake)
- observe how check_compliance.py -m Codeowners now takes minutes
- create the following symbolic link: ln -s ../zephyr myself
- observe how check_compliance.py -m Codeowners now takes hours or more.

Only "hours" and not infinity because operating systems have limits to
break infinite loops of symbolic links.

Now considering Zephyr supports Windows and none of the multiple flavors
of symbolic links offered by Windows is usable, it's unlikely Zephyr
will ever _require_ symbolic links. However this shouldn't _forbid_
other users to use symbolic links privately.

Funny enough neither glob nor pathlib documents whether they follow
symbolic links to directories in recursive mode. "glob" does and
"pathlib" doesn't. So I thought this was bug in glob and even had a
tentative fix for it. Then I scanned the glob's git log and found it's
actually a "feature":
https://github.com/python/cpython/pull/12918. glob's own documentation
mentions pathlib as a higher-level and newer alternative: let's switch
to pathlib.

Other changes:

- Fixed failing "path not found" detection of bogus lines in
  CODEOWNERS. glob seems to have a bug where glob.glob('doesntexist/**',
  recursive=True) returns 'doesntexit/' instead of '' (easter egg: there
  is one bogus line right now)

- Successfully tested on Windows. Requires hack replacing sh with pbs
  and disabling a couple unrelated imports though. Previous glob version
  failed to match, new files were never owned.

- Remove (accidental?) limitation to run from the git top level. Moved
  git top-level command to a global and remove the same limitation on -m
  Checkpatch too.

- Tighten the regular expression parsing CODEOWNERS lines to actually
  report invalid lines: make sure pattern starts in first character;
  make sure the owner is not mister whitespace.

Performance:

- Down from 1 sec to less than 0.5s in the most common case: not adding
  any new file. Fix broken optimization that still scanned the whole
  tree even when there are no new files (hint: [''] is not false)

- About 20% faster on large sets without symbolic links (up to
  "infinitely" faster *with* symlinks) One possible reason: the list of
  owned files is now split per CODEOWNERS line for better logging which
  also means scanning smaller lists.

Other minor fixes:

- Remove unused git ls-files invocation, replaced with an optimization
  TODO.
- Use more portable splitlines() instead of split('\n')
- renamed parse_codeowners() to ls_owned_files() because but it does far
  more than parsing CODEOWNERS
- Use "early returns" and extract new function to reduce cyclomatic
  complexity (up to 9 levels of indentation before in
  parse_codeowners(). Other minor simplifications.
- Add some pydocs and logging

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -19,8 +19,7 @@ from shutil import copyfile
 import json
 import tempfile
 from colorama import Fore, Style
-import glob
-import yaml
+from pathlib import Path
 
 logger = None
 
@@ -36,6 +35,8 @@ sh_special_args = {
     '_cwd': os.getcwd()
 }
 
+# The absolute path of the top-level git directory
+git_top = sh.git("rev-parse", "--show-toplevel").strip()
 
 def get_shas(refspec):
     """
@@ -73,7 +74,6 @@ class ComplianceTest:
         self.case = None
         self.suite = suite
         self.commit_range = commit_range
-        self.repo_path = os.getcwd()
         # get() defaults to None if not present
         self.zephyr_base = os.environ.get('ZEPHYR_BASE')
 
@@ -163,7 +163,7 @@ class CheckPatch(ComplianceTest):
     def run(self):
         self.prepare()
         # Default to Zephyr's checkpatch if ZEPHYR_BASE is set
-        checkpatch = os.path.join(self.zephyr_base or self.repo_path, 'scripts',
+        checkpatch = os.path.join(self.zephyr_base or git_top, 'scripts',
                                   'checkpatch.pl')
         if not os.path.exists(checkpatch):
             self.skip(checkpatch + " not found")
@@ -173,7 +173,8 @@ class CheckPatch(ComplianceTest):
         try:
             subprocess.check_output((checkpatch, '--mailback', '--no-tree', '-'),
                                     stdin=diff.stdout,
-                                    stderr=subprocess.STDOUT, shell=True)
+                                    stderr=subprocess.STDOUT,
+                                    shell=True, cwd=git_top)
 
         except subprocess.CalledProcessError as ex:
             output = ex.output.decode("utf-8")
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -172,7 +172,7 @@ def run(self):
                                 stdout=subprocess.PIPE)
         try:
             subprocess.check_output((checkpatch, '--mailback', '--no-tree', '-'),
-                                    stdin=diff.stdout,
+                                    stdin=diff.stdout, cwd=git_top,
```
ulfalizer: Bit odd to have `cwd` sandwiches between `stdin` and `stderr`.

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
     _name = "Codeowners"
     _doc  = "https://help.github.com/articles/about-code-owners/"
 
-    def parse_codeowners(self, git_root, codeowners):
-        all_files = []
+
+    def ls_owned_files(self, codeowners):
+        """Returns a OrderedDict mapping git patterns from the CODEOWNERS file
```
ulfalizer: *an

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+        # TODO: filter out files not in "git ls-files" (e.g.,
+        # sanity-out) _if_ the overhead isn't too high for a clean tree.
+        #
+        # pathlib.match() doesn't support **, so it looks like we can't
+        # recursively glob the output of ls-files directly, only real
+        # files :-(
+
+        pattern2files = collections.OrderedDict()
+        top_path = Path(git_top)
+        top_len = len(git_top) + 1 # 1 for the slash
```
ulfalizer: Style nit: Usually two spaces before `#` comments after line of code

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
-                            all_files += files
-
-        files = []
-        for f in all_files:
-            if os.path.isfile(f):
-                files.append(f)
-
-        return set(files)
+            linenum = 0
+            for line in codeo:
```
ulfalizer: Can do `for lineno, line in enumerate(codeo, 1):`.

(I used `linenr` in Kconfiglib, but later on realized that `lineno` is the "standard" in a bunch of other places, so just to avoid repeating that. :P)

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+                        "Invalid CODEOWNERS line %d\n\t%s" %
+                        (linenum, line))
+                    continue
+
+                git_patrn = match.group(1)
+                glob = self.git_pattern_to_glob(git_patrn)
+                files = []
+                for p in top_path.glob(glob):
+                    # comparing strings is much faster later
+                    files.append(str(p)[top_len:])
```
ulfalizer: `files = [str(p)[top_len:] for p in top_path.glob(glob)]` maybe.

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+
+                if not files:
+                    self.add_failure("Path '{}' not found, in "
+                                     "CODEOWNERS".format(git_patrn))
+
+                pattern2files[git_patrn] = files
+
+        return pattern2files
+
+
```
ulfalizer: Only one blank line between methods usually (though it's a bit hypocritical for me to complain, because I used to put a single blank line between everything... just to make it consistent :P).

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+        """Appends and prepends '**[/*]' when needed. Result has neither a
+        leading nor a trailing slash.
+        """
+
+        ret = git_pattern
+
+        if git_pattern.startswith("/"):
+            ret = ret[1:]
+        else:
+            ret = "**/" + ret
```
ulfalizer: Get rid of `ret = git_pattern` and do this maybe:

```python
if git_pattern.startswith("/"):
    ret = git_pattern[1:]
else:
    ret = "**/" + git_pattern
```

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
-        git_root = sh.git("rev-parse", "--show-toplevel").strip()
-        codeowners = os.path.join(git_root, "CODEOWNERS")
+        # TODO: testing an old self.commit range that doesn't end
+        # with HEAD is most likely a mistake. Should warn, see
+        # https://github.com/zephyrproject-rtos/ci-tools/pull/24
+        codeowners = os.path.join(git_top, "CODEOWNERS")
         if not os.path.exists(codeowners):
             self.skip("CODEOWNERS not available in this repo")
 
         commit = sh.git("diff","--name-only", "--diff-filter=A", self.commit_range, **sh_special_args)
```
ulfalizer: Missing space after `,`.

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+
+        if not new_files:
+            # TODO: parse CODEOWNERS to report errors in it *without*
+            # scanning the tree to keep this case very fast.
+            return
+
+        # Convert to pathlib.Path string representation (e.g.,
+        # double-backslashes 'dir1\\dir2\\' on Windows) to be consistent
+        # with self.ls_owned_files()
+        new_files = [ str(Path(f)) for f in new_files ]
```
ulfalizer: Usually no space after `[` and before `]`.

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+
+        logging.info("If this takes too long then cleanup and try again")
+        patrn2files = self.ls_owned_files(codeowners)
+
+        new_not_owned = []
+        for f in new_files:
```
ulfalizer: Could call it `new_file` instead of `f` maybe.

scripts/check_compliance.py
```diff
@@ -426,73 +426,123 @@ class Codeowners(ComplianceTest):
...
+            f_is_owned = False
+
+            for git_pat, owned in patrn2files.items():
+                logging.debug("Scanning %s for %s" % (git_pat, f))
+
+                if f in owned:
+                    logging.info("%s matches new file %s", git_pat, f)
+                    f_is_owned = True
+                    break
```
ulfalizer: Maybe it's harder to read for people who don't do that much Python, but another option is to use a `for ... else`:

```python
for git_pat, owned in patrn2files.items():
    ...
    if new_file in owned:
        ...
        break
else:
    new_not_owned.append(new_file)
```

scripts/check_compliance.py
```diff
@@ -36,6 +36,7 @@ def error(what):
     '_cwd': os.getcwd()
 }
 
+git_top = sh.git("rev-parse", "--show-toplevel").strip()
```
ulfalizer: Could add a comment above this maybe. "The absolute path yada yada..."

scripts/check_compliance.py
ulfalizer: And `p` -> `path` maybe. I'm horrible at reading single-letter variable names, even when they're obvious. I could never go into math. ;)

scripts/check_compliance.py
marc-hb: I admittedly didn't run pylint/pep8 yet, will do.

scripts/check_compliance.py
marc-hb: ?

scripts/check_compliance.py
marc-hb: (Pointless) effort to minimize the diff size; will re-order.

scripts/check_compliance.py
marc-hb: Please let me not put my name on the git blame of at least this line :-)

scripts/check_compliance.py
marc-hb: Agree this is a perfect use case for this "for/else" syntax. Also agree this syntax is very unusual, hence harder to read and would require almost as much comment as the amount of code it would save. Guido wrote he regrets this feature: https://intoli.com/blog/for-else-in-python/
Any chance it would help with performance? If not then I'd rather not.

scripts/check_compliance.py
ulfalizer: "an OrderedDict" instead of "a OrderedDict", since 'O' is a vowel. :)

scripts/check_compliance.py
ulfalizer: Fine with me. I have a silly obsession with scrunching everything up. :P

scripts/check_compliance.py
ulfalizer: Ohh, sorry, thought it was part of the diff.

scripts/check_compliance.py
ulfalizer: Fine with me to skip it.

marc-hb:
This check is designed to catch new files only. However by changing just one line it's trivial to list all orphaned files already checked in. Here are the directories with the most orphaned files right now:
```
soc/arc/  # clone arch/arc/ owners?
arch/x86_64/
soc/x86_64/
lib/cmsis_rtos_v1,2/
lib/libc/
dts/arm/<several>
scripts/<many>
drivers/<many>
subsys/<many>
```
Also:
```
include/
tests/
samples/
ext/ # going away
```

```diff
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -503,7 +503,7 @@ class Codeowners(ComplianceTest):
         if not os.path.exists(codeowners):
             self.skip("CODEOWNERS not available in this repo")
 
-        commit = sh.git("diff","--name-only", "--diff-filter=A", self.commit_range, **sh_special_args)
+        commit = sh.git("ls-files", **sh_special_args)
         new_files = commit.splitlines()
         logging.debug("New files %s", new_files)
```

PR #55, opened 2019-05-01
check_compliance.py: Exclude simple Kconfig token pasting

Sizurka:
Alter the regular expression used in git-grep to exclude simple token pasting for the form `CONFIG_STUFF##something`.  This does not catch any use for `UTIL_CAT` and friends, but it does allow removal of some elements from the whitelist.

The specific motivation is zephyrproject-rtos/zephyr#15105 where I used token pasting for counter instances.  However, this is a general solution to the simple cases like that.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -310,15 +310,22 @@ entries, then bump the 'max_top_items' variable in {}.
         #
         #   foo/bar.c<null>17<null>CONFIG_BAZ
         #
+        # The regular expression in use here uses zero length word boundary
+        # assertions (\b) to isolate the reference then a negative lookahead
+        # to ensure that it is not followed by a token paste operation (##).
+        # So this excludes the root of any references used as part of token
+        # pasting, since they would always need to be whitelisted anyway.
+        #
         # Skip the samples/ and tests/ directories for now. They often contain
         # Kconfig files that are not part of the main Kconfig tree, which will
         # trigger false positives until we do something fancier. Skip
         # doc/releases too, which often references removed symbols.
-        grep_cmd = "git grep --only-matching --line-number -I --null " \
-                   "--extended-regexp --word-regexp CONFIG_[A-Z0-9_]+ " \
-                   "-- :!samples :!tests :!doc/releases"
+        grep_cmd = ["git", "grep", "--only-matching", "--line-number", "-I",
+                    "--null", "--perl-regexp",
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -314,11 +314,12 @@ def check_no_undef_outside_kconfig(self, kconf):
         # Kconfig files that are not part of the main Kconfig tree, which will
         # trigger false positives until we do something fancier. Skip
         # doc/releases too, which often references removed symbols.
-        grep_cmd = "git grep --only-matching --line-number -I --null " \
-                   "--extended-regexp --word-regexp CONFIG_[A-Z0-9_]+ " \
-                   "-- :!samples :!tests :!doc/releases"
+        grep_cmd = ["git", "grep", "--only-matching", "--line-number", "-I",
+                    "--null", "--perl-regexp",
+                    r"\bCONFIG_[A-Z0-9_]+\b(?!\s*#)",
```
ulfalizer: Is there any harm in making this `(?!\s*##)` by the way, just to make it a bit more specific?

The check searches comments and stuff too, where maybe `CONFIG_FOO  #...` could show up in some context, though it's unlikely.

Sizurka: Done for both.  I also missed an instance that could be removed from the whitelist the first time through (`CONFIG_UART_`); I only checked for ones that matched the new exclusion (i.e. that where present as simple token pastes), but I managed to miss that one.

ulfalizer: LGTM

galak:
While not related to this PR, what to people feel about moving the whitelist into a file and keeping it in the zephyr repo itself.

ulfalizer:
@galak 
The reason I didn't do that right away is that I think the right solution is to move `ci-tools` back into the main Zephyr repo. There will be other whitelists later on too, and it's annoying if you have to have a custom whitelist file for each, that supports comments, etc., and with the extra code that comes with that.

It feels like working around that `ci-tools` would be better to have as `scripts/ci` in the Zephyr repo, where it used to be.

I was kinda hoping that someone would take the initiative and move it back, but I could make a PR for it later.

ulfalizer:
@Sizurka 
Spottedanother thing: The comment above `UNDEF_KCONFIG_WHITELIST` needs to be updated. Could just say "many of these are symbols used as example" now.

Sizurka:
Ok, comment updated.

PR #57, opened 2019-05-01
check_compliance.py: show which (sub)directory is being checked

marc-hb:
When run outside ZEPHYR_BASE, the tests in check_compliance.py behave in
various ways: some check the current repo while others stick to
ZEPHYR_BASE. Some run in a subdirectory while others stick to GIT_TOP,
Some don't mind being run from a subdirectory while others fail.

Some of these inconsistencies make sense while others are limitations
that should probably be fixed. However the first step is awareness and
clarifying mysteries so this commit purely prints which (sub)directory
is being checked for every test while making absolutely zero functional
change.

Sample output:
```
cd zephyr-ci-tools/scripts/

Running checkpatch       in /home/joe/zephyr-ci-tools ...
Running Kconfig          in /home/joe/zephyr ...
Running Codeowners       in /home/joe/zephyr-ci-tools ...
Running Documentation    in /home/joe/zephyr-ci-tools/scripts ...
Running Gitlint          in /home/joe/zephyr-ci-tools/scripts/[.gitlint]..
Running License          in /home/joe/zephyr-ci-tools/scripts ...
Running Identity/Emails  in /home/joe/zephyr-ci-tools ...
```
get ZEPHYR_BASE env only once, not ~ 15 times

Also rename git_top constant to GIT_TOP because pylint says so.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -78,14 +78,14 @@ class ComplianceTest:
         self.commit_range = commit_range
         # get() defaults to None if not present
 
-    def prepare(self):
+    def prepare(self, where):
         """
         Prepare test case
         :return:
         """
         self.case = MyCase(self._name)
         self.case.classname = "Guidelines"
-        print("Running {} tests...".format(self._name))
+        print("Running {:16} tests in {} ...".format(self._name, where))
 
     def run(self):
         """
@@ -162,13 +162,14 @@ class CheckPatch(ComplianceTest):
     _doc = "https://docs.zephyrproject.org/latest/contribute/#coding-style"
 
     def run(self):
-        self.prepare()
+        self.prepare(GIT_TOP)
         # Default to Zephyr's checkpatch if ZEPHYR_BASE is set
         checkpatch = os.path.join(ZEPHYR_BASE or GIT_TOP, 'scripts',
                                   'checkpatch.pl')
         if not os.path.exists(checkpatch):
             self.skip(checkpatch + " not found")
 
+        # git diff's output doesn't depend on the current (sub)directory
         diff = subprocess.Popen(('git', 'diff', '%s' % (self.commit_range)),
                                 stdout=subprocess.PIPE)
         try:
@@ -192,7 +193,7 @@ class KconfigCheck(ComplianceTest):
     _doc = "https://docs.zephyrproject.org/latest/tools/kconfig/index.html"
 
     def run(self):
-        self.prepare()
+        self.prepare(ZEPHYR_BASE)
 
         kconf = self.parse_kconfig()
 
@@ -556,7 +557,7 @@ class Documentation(ComplianceTest):
     DOCS_WARNING_FILE = "doc.warnings"
 
     def run(self):
-        self.prepare()
+        self.prepare(os.getcwd())
 
         if os.path.exists(self.DOCS_WARNING_FILE) and os.path.getsize(self.DOCS_WARNING_FILE) > 0:
             with open(self.DOCS_WARNING_FILE, "rb") as docs_warning:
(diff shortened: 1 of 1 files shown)


marc-hb:
```
Running Gitlint          in /home/joe/zephyr-ci-tools/scripts/[.gitlint]..
```
perfectly proving the point of this PR: the `.gitlint` file is of course not in the `scripts/` subdirectory, which is exactly why I missed this:
> Gitlint issues Commit 05831c8:
> 20: UC4 Line exceeds max length (74>72): "Running Gitlint in /home/joe/zephyr-ci-> tools/scripts/[.gitlint].."

scripts/check_compliance.py
```diff
@@ -571,7 +573,7 @@ class GitLint(ComplianceTest):
     _doc = "https://docs.zephyrproject.org/latest/contribute/#commit-guidelines"
 
     def run(self):
-        self.prepare()
+        self.prepare(os.path.join(os.getcwd(), '[.gitlint]'))
```
mbolivar: What's with the brackets?

scripts/check_compliance.py
```diff
@@ -35,8 +35,10 @@ def error(what):
     '_cwd': os.getcwd()
 }
 
+ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE')
```
mbolivar: `os.environ['ZEPHYR_BASE']` would crash if unset instead of returning `None` like `get` does. That's more fail-fast.

marc-hb: Almost all the checks work without any ZEPHYR_BASE (the one that doesn't prints "None")

scripts/check_compliance.py
marc-hb: It means "optional" like in man pages. Any better idea?

mbolivar: Now I'm even more confused... so the output says this:

```
Running Gitlint          in /home/joe/zephyr-ci-tools/scripts/[.gitlint]..
```

Does that mean it looks for both `/home/joe/zephyr-ci-tools/scripts/[.gitlint]` and `/home/joe/zephyr-ci-tools/scripts`?

scripts/check_compliance.py
mbolivar: That's surprising! Thanks for explaining.

scripts/check_compliance.py
marc-hb: The GitLint check looks for just that `.gitlint` file and falls back on defaults if not found. So if you run it from a subdirectory like I did then it runs "successfully" except with the default settings instead of the repo settings. See example above. This should a one-line fix but later; I really don't want to bury/hide any fix now in this relatively large but zero functional change. So the main purpose is to draw the user's attention/suspicion on the hidden `.gitlint` file and it seems it succeeded in at least that :-) Before I wrote this patch I had completely forgotten about this hidden file myself! Then I didn't expect another .gitlint to be found in zephyr-ci-tools/ in addition to the one in ZEPHYR_BASE/.

So yes, I agree it's struggling to cram a bit too much information in a very limited space. On the other hand the entire `class GitLint` is 20 lines long including comments if anyone wants to get to the bottom of this.

mbolivar: 🤷‍♂  i find this confusing but I don't think it should block the merge.

marc-hb: Thanks, I promise to submit the one-line change that makes gitlint more robust once this is merged (and proved not to cause any problem for a few days)

ulfalizer: Could spell it out maybe, like

```
Running Gitlint          in /home/joe/zephyr-ci-tools/scripts (possibly with settings from a .gitlint file in that directory)
```

Or the code could check if there's actually a `.gitlint` in there and alter the message.

marc-hb: As promised, gitlint fix submitted at PR #60

PR #62, opened 2019-05-09
check_compliance.py: Add Git helper and get rid of sh_special_args

ulfalizer:
The sh library (https://amoffat.github.io/sh/) is only used to run Git,
and each call to sh.git() currently passes _tty_out=False to prevent Git
from starting a pager (through sh_special_args).

Add comments to explain what 'sh' is, why _tty_out=False is there, and
add a helper function for running Git. That way the mysterious
sh_special_args dictionary can be removed as well.

Importing a library just for running Git might be overkill, but this
should immediately demystify what's going on at least.

Another issue with the 'sh' library is that it doesn't work on Windows.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -23,34 +23,42 @@ from pathlib import Path
 
 logger = None
 
+
 def info(what):
     sys.stdout.write(what + "\n")
     sys.stdout.flush()
 
+
 def error(what):
     sys.stderr.write(Fore.RED + what + Style.RESET_ALL + "\n")
 
-sh_special_args = {
-    '_tty_out': False,
-    '_cwd': os.getcwd()
-}
+
+def git(*args):
+    # Runs a git command using the 'sh' library (https://amoffat.github.io/sh/)
+
+    # Do not create a TTY for stdout, so that Git doesn't start a pager.
+    #
+    # Hack: Setting _cwd to the working dir seems pointless as of writing (it
+    # should be the default), but keep it around in case it's working around
+    # some issue
+    return sh.git(*args, _tty_out=False, _cwd=os.getcwd())
+
 
 ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE')
 # The absolute path of the top-level git directory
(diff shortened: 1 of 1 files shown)


ulfalizer:
Could tack on `_tty_out=False` to all the calls and get rid of the helper too. I'm fine with either one.

ulfalizer:
> Another good rationale you can add in the commit message: the sh module doesn't work on Windows; it's the cause of the biggest hack (Hi, stackoverflow!) I needed to run check_compliance.py on Windows. Encapsulating it is a great step towards getting rid of it.

Added Windows compatibility note.

> Out of caution please keep the (also weird) _cwd argument though. It costs extremely few characters and God knows what bug in this unmaintained python module it was meant to work around or what undocumented behaviour change it could cause. Plus it'll make this an absolute "zero functional change" commit; guaranteed.

Would be nice to get rid of just to prevent just-in-case'ing in the future, if it's not needed.

@nashif 
Do you remember why the `_cwd=os.getcwd()` argument was added?

marc-hb:
> Would be nice to get rid of just to prevent just-in-case'ing in the future, if it's not needed.

I saw the future and "sh" wasn't part of it :-) Leaving it is a no-op operation: just-in-case before, just-in-case now.

Can you at least move it to a separate commit for a better git log and easier revert if needed?

ulfalizer:
Added back the `_cwd=os.getcwd()`.

Still a functional change in that it always used the `os.getcwd()` at startup previously though. Could work around that by saving `os.getcwd()` at startup or the like, but might be a bit paranoid.

marc-hb:
> Added back the _cwd=os.getcwd().

Actually: the commit message still says it's gone.

(I really don't like github's efforts to hide commit messages and the disconnection with the "conversation top" even more. I'm used to them being first class review citizens)

ulfalizer:
> Actually: the commit message still says it's gone.

Whoops, fixed.

ulfalizer:
@nashif 
Remember why the `_cwd` was added?

PR #65, opened 2019-05-17
check_compliance.py: Avoid more false positives in Kconfig check

ulfalizer:
Add more patterns to the negative lookahead in the regex, to avoid
flagging CMake variable expansion (`CONFIG_FOO_${VAR}`), CMake
`configure_file()` references (`CONFIG_FOO_@VAR@`), Python `format()` stuff
(`"CONFIG_FOO_{}_...".format(...)`), and `#endif` comments that talk about
many symbols (`#endif /* CONFIG_FOO_* */`).

This gets rid of some entries from the whitelist too.

Mention the safe #endif format in the message posted when the test
fails.

Prompted by https://github.com/zephyrproject-rtos/zephyr/pull/16222.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -321,10 +321,13 @@ entries, then bump the 'max_top_items' variable in {}.
         #   foo/bar.c<null>17<null>CONFIG_BAZ
         #
         # The regular expression in use here uses zero length word boundary
-        # assertions (\b) to isolate the reference then a negative lookahead
-        # to ensure that it is not followed by a token paste operation (##).
-        # So this excludes the root of any references used as part of token
-        # pasting, since they would always need to be whitelisted anyway.
+        # assertions (\b) to isolate the reference then a negative lookahead to
+        # ensure that it is not followed by a token paste operation
+        # (CONFIG_FOO_##...), a '$' (e.g. for a CMake variable expansion like
+        # CONFIG_FOO_${VAR}), a '@' (for a CMake configure_file() reference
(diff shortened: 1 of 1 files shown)


ulfalizer:
CC @Sizurka

Sizurka:
You may want to explicitly escape the `{` in the lookahead, since that's also also a reserved character.  The git parser is apparently smart enough to pass it through, since there it's not of the form `{N}` or `{N,M}` (either as repetition of a modifier).  But that's not guaranteed behavior (i.e. Perl itself will throw a syntax error if you try that).

ulfalizer:
@Sizurka 
Fixed.

I broadened it a bit to just look for `$` instead of `${`, and switched to a character set. Probably unlikely that `CONFIG_FOO` is referenced as `CONFIG_FOO$` (though it might for super-obscure cases with shell scripts maybe...).

ulfalizer:
@marc-hb 
I still think the regexp is quicker to understand and tweak than doing it manually in Python, even if it was without comments (plus it's neater to do it in one place). I just like commenting a lot.

Most of the comments explain why those particular characters were picked too. Would still get those with Python I guess.

ulfalizer:
Hmm... think you'd also need to capture some context around the match to implement the same thing in Python. That offsets it more, IMO.

ulfalizer:
@marc-hb 
Are you okay with merging this version? Would be nice so that https://github.com/zephyrproject-rtos/zephyr/pull/16222 can go in.

marc-hb:
> Are you okay with merging this version?

For the record yes I was, I would have "requested changes" otherwise. I didn't see it was blocking the other PR sorry.

PR #68, opened 2019-05-31
check_compliance.py: Avoid using newly-added 'git grep --only-matching'

ulfalizer:
--only-matching was added to 'git grep' during the second half of 2018,
so relying on it is shaky. Rewrite check_no_undef_outside_kconfig() to
manually extract the symbol reference(s) from each matching line
instead.

Also fix an issue with a literal '{' in a string passed to .format()
that raises an error. Needs to be doubled.

The format of the Kconfig whitelist was changed to omit the CONFIG_
prefix from each entry.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -318,7 +318,11 @@ entries, then bump the 'max_top_items' variable in {}.
         # Example output line for a reference to CONFIG_BAZ at line 17 of
         # foo/bar.c:
         #
-        #   foo/bar.c<null>17<null>CONFIG_BAZ
+        #   foo/bar.c<null>17<null>#ifdef CONFIG_BAZ
+        #
+        # 'git grep --only-matching' would get rid of the surrounding context
+        # ('#ifdef '), but it was added fairly recently (second half of 2018),
+        # so we extract the references from each line ourselves instead.
         #
         # The regular expression in use here uses zero length word boundary
         # assertions (\b) to isolate the reference then a negative lookahead to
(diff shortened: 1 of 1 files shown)


ulfalizer:
CC @nashif @carlescufi @marc-hb 

The lookahead needs to be checked twice since you might get a line like

    CONFIG_FOO yada yada CONFIG_BAR_##x

The Python code needs to only extract `CONFIG_FOO` there as well.

scripts/check_compliance.py
```diff
@@ -333,9 +337,12 @@ def check_no_undef_outside_kconfig(self, kconf):
         # Kconfig files that are not part of the main Kconfig tree, which will
         # trigger false positives until we do something fancier. Skip
         # doc/releases too, which often references removed symbols.
-        grep_cmd = ["git", "grep", "--only-matching", "--line-number", "-I",
-                    "--null", "--perl-regexp",
-                    r"\bCONFIG_[A-Z0-9_]+\b(?!\s*##|[$@{*])",
+
```
marc-hb: > // Those would always need to be whitelisted anyway.

While we're here could you explain this (older) comment above? Who are "those"?

scripts/check_compliance.py
```diff
@@ -356,14 +363,17 @@ def check_no_undef_outside_kconfig(self, kconf):
...
-            path, lineno, sym_name = line.split("\0")
+        for grep_line in grep_stdout.decode("utf-8").splitlines():
+            path, lineno, line = grep_line.split("\0")
 
-            # [7:] removes the "CONFIG_" prefix
-            if sym_name[7:] not in defined_syms and \
-               sym_name not in UNDEF_KCONFIG_WHITELIST:
+            # Extract symbol references (might be more than one) within the
+            # line
+            for sym_name in re.findall(regex, line):
```
marc-hb: Did you manually test multiple matches on the same line? Just to be on the safe side. I guess these must be rare.

ulfalizer: Yup, accounting for multiple matches on one line.

scripts/check_compliance.py
ulfalizer: Piggy-backed a separate commit that cleans up that comment.

I just removed that sentence. Was trying to say that if there was no negative lookahead, then those would always have to be (manually) whitelisted.

PR #69, opened 2019-06-07
set labels on PR

nashif:
Look at files being changed and set labels accordingly.

### DIFF
--- a/scripts/what_changed.py
+++ b/scripts/what_changed.py
@@ -1,0 +1,344 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# A script to set labels on pull-rquests based on files being changed
+
+import sys
+import re, os
+from email.utils import parseaddr
+import sh
+import argparse
+from github import Github
+
+if "ZEPHYR_BASE" not in os.environ:
+    exit(1)
+
+repository_path = os.environ['ZEPHYR_BASE']
+sh_special_args = {
+    '_tty_out': False,
+    '_cwd': repository_path
+}
+
+def git(*args):
+    # Runs a git command using the 'sh' library (https://amoffat.github.io/sh/)
+
+    # Do not create a TTY for stdout, so that Git doesn't start a pager.
+    #
+    # Hack: Setting _cwd to the working dir seems pointless as of writing (it
+    # should be the default), but keep it around in case it's working around
+    # some issue
+    return sh.git(*args, _tty_out=False, _cwd=os.getcwd())
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+                description="Set labels for a pull request based on files that were changed")
+    parser.add_argument('-c', '--commits', default=None,
+            help="Commit range in the form: a..b")
+    parser.add_argument('-r', '--repo', default=None,
+                        help="Github repository")
+    parser.add_argument('-p', '--pull-request', default=0, type=int,
+                        help="Pull request number")
+
+    return parser.parse_args()
+
+def main():
+    boards = set()
+
+    args = parse_args()
+    if not args.commits:
+        exit(1)
+
+    commit = git("diff","--name-only", args.commits)
+    files = commit.split("\n")
+
+    docs = 0
+    images = 0
+    kernel = 0
+    code = 0
+    net = 0
+    bt = 0
+
+    matches = [
+            {
+                "area": "Modem",
+                "regex" : ["^drivers/modem"],
+                "counter": 0,
+                "labels": ["area: Modem"]
+                },
+            {
+                "area": "PWM",
+                "regex" : ["^drivers/pwm"],
+                "counter": 0,
+                "labels": ["area: PWM"]
+                },
+            {
+                "area": "C Library",
+                "regex" : ["^lib/libc"],
+                "counter": 0,
+                "labels": ["area: C Library"]
+                },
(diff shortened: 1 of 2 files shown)


galak:
> Look at files being changed and set labels accordingly.

YEAH!!!!

scripts/what_changed.py
```diff
@@ -0,0 +1,341 @@
...
+    files = commit.split("\n")
+
+    docs = 0
+    images = 0
+    kernel = 0
+    code = 0
+    net = 0
+    bt = 0
+
+    matches = [
```
galak: Is it possible that we encode the 'matches' as a file in the gitrepo similar to CODEOWNERS or MAINTAINER file syntax so we can update it as new code area/labels are created?

nashif: yeah, thought about it, I now want to get this in, get some feedback, and optimise. But a label file similar to CODEOWNER is definitely on the plan.

I am also looking for a way to identify complexity of changes and set either Maintainer or Trivial labels.

galak: Agreed, having something working first will be great.  Can't wait for API, DTS labels to get automatically set!!

nashif:
example of labels set by script:

https://github.com/zephyrproject-rtos/zephyr/pull/16662

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+                },
+            {
+                "area": "POSIX",
+                "regex" : ["^lib/posix"],
+                "counter": 0,
+                "labels": ["area: POSIX"]
+                },
+            {
+                "area": "Native Port",
+                "regex" : ["^arch/posix"],
```
aescolar: Could you add to this `*native_posix*` and `^soc/posix`

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+                },
+            {
+                "area": "Storage",
+                "regex" : ["^subsys/storage"],
+                "counter": 0,
+                "labels": ["area: Storage"]
+                },
+            {
+                "area": "Bluetooth",
+                "regex" : ["^subsys/bluetooth"],
```
aescolar: could you change this one to just `*bluetooth*` to also catch their samples and tests?

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+                },
+            {
+                "area": "Documentation",
+                "regex" : ["^doc/", "\.rst$", "\.txt$"],
+                "counter": 0,
+                "labels": ["area: Documentation"]
+                },
+            {
+                "area": "Build System",
+                "regex" : ["^cmake/", "^CMakeLists.txt", "\.cmake"],
```
carlescufi: the last one should be removed. There are lots of `.cmake` files in the `boards/` folder but this should not set the "Build System" label

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+                },
+            {
+                "area": "SPI",
+                "regex" : ["^drivers/spi"],
+                "counter": 0,
+                "labels": ["area: SPI"]
+                },
+            {
+                "area": "Boards",
+                "regex" : ["^board"],
```
carlescufi: should be `^boards`?

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# A script to generate a list of boards that have changed or added and create an
```
aescolar: This does not seem correct :)

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+    #
+    # Hack: Setting _cwd to the working dir seems pointless as of writing (it
+    # should be the default), but keep it around in case it's working around
+    # some issue
+    return sh.git(*args, _tty_out=False, _cwd=os.getcwd())
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+                description="Generate a sanitycheck argument for for boards "
```
aescolar: This does not seem correct :)

scripts/what_changed.py
carlescufi: Agreed, let's have a second pass with the regexes and labels moved to a separate file

scripts/what_changed.py
carlescufi: For example the PR you linked should not have a "Build System" label IMHO: https://github.com/zephyrproject-rtos/zephyr/pull/16662/

scripts/what_changed.py
```diff
@@ -0,0 +1,345 @@
...
+    kernel = 0
+    code = 0
+    net = 0
+    bt = 0
+
+    matches = [
+            {
+                "area": "Modem",
+                "regex" : ["^drivers/modem"],
+                "counter": 0,
```
carlescufi: just place the counters in a separate array of ints that is the same length as matches, then you have no need to add this to every single entry

carlescufi: Ah but you'll need `OrderedDict` then. Up to you, you could also create a simple container class `Area` that has `name`, `regex`, `counter`, and `labels` as members

nashif: that is script i wrote like a year ago very fast and just brought it back to life, I have no intention to keep this structure and maintain it here, idea is to move to the tree similar to CODEOWNERS later, so all of this will be scrapped.

PR #72, opened 2019-07-04
codeowner: do not allow commas

nashif:
Do not allow comma in path, this is not allowed by Github parser.


aescolar:
Thanks for that

PR #73, opened 2019-07-23
Run the dtlib and edtlib test suites in CI

ulfalizer:
zephyrproject-rtos/zephyr#17660 must go in
before this will work.

Piggyback a clarification re. ZEPHYR_BASE being None when not running on
a Zephyr tree (e.g. when running on the ci-tools repo itself).

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -44,7 +44,9 @@ def git(*args):
     return sh.git(*args, _tty_out=False, _cwd=os.getcwd())
 
 
+# This ends up as None when we're not running in a Zephyr tree
 ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE')
+
 # The absolute path of the top-level git directory
 GIT_TOP = git("rev-parse", "--show-toplevel").strip()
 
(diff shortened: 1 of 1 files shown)


ulfalizer:
@galak 
Yep, but it runs in a fraction of a second, so it's not an issue.

carlescufi:
Is this something that should be in the  compliance check or rather in CI itself, like the west unit tests?

ulfalizer:
@carlescufi 
What do you mean by "in CI itself"? This is the only way of running tests that I know about. Do we need several?

It's testing dtlib and edtlib lib themselves. The important thing to me is that a message shows up on GitHub if you break them. Hiding it in a log would be wasting people's time.

ulfalizer:
I'll probably add global checks for the Zephyr code later as well, like for Kconfig, if it's the word "compliance" that bothers you (broken code is non-compliant).

Regardless of what we call them, this type of test is the right place to do it, to make things nicer for users.

carlescufi:
@galak can you please take a look and merge if you have no objections?

galak:
Merging, but we need to clean this up in the future so this isn't run on every PR.

PR #79, opened 2019-09-11
Report exceptions

ulfalizer:


scripts/check_compliance.py
```diff
@@ -895,211 +891,225 @@ def init_logs(cli_arg):
...
-    :return:
-    """
+    global commit_sha
+    global github_repo
+    global github_pr
+    global github_commit
+
+    if not (args.github or args.status):
+        # No GitHub-related stuff to do
+        return
```
marc-hb: I think it would be slightly more readable to have this test in main().

Extracting github stuff most users never run into a different script would also make this one easier to read and maintain however. 1. I don't if that's realistic 2. I digress.

scripts/check_compliance.py
```diff
@@ -895,211 +891,225 @@ def init_logs(cli_arg):
...
-    comment_count = 0
+    name2doc = {testcase._name: testcase._doc
+                for testcase in ComplianceTest.__subclasses__()}
 
-    print("Processing results...")
+    def set_commit_status(status, msg):
+        # 'case' gets set in the loop.
+        # pylint: disable=undefined-loop-variable
+        github_commit.create_status(status, name2doc[case.name], msg,
+                                    case.name)
```
marc-hb: I'm curious whether this requires special privileges or can any user (accidentally or not) override CI status. Good for testing this script, maybe not as good for creating github confusion :-)
Only for their own commits maybe?

scripts/check_compliance.py
ulfalizer: Yeah, could change it later maybe.

PR #80, opened 2019-09-11
Be alive

ulfalizer:
We need to fix our CI so that it doesn't just silently do nothing on most errors.


mbolivar:
What exactly is the problem? Why are these patches being reverted?

ulfalizer:
> What exactly is the problem?

CI isn't running. It usually does on this repo too: https://github.com/zephyrproject-rtos/ci-tools/pull/37#issuecomment-474979313

Seeing if it's something that was added recently that broke it.

ulfalizer:
@mbolivar 
Did anything west-related change recently that might've broken this?

ulfalizer:
It just posted a message on https://github.com/zephyrproject-rtos/ci-tools/pull/79. I wonder if it's just being extremely slow.

PR #89, opened 2019-09-16
what_changed.py: Remove unused variables and imports

ulfalizer:
Fixes pylint warnings and makes the script easier to read.

Also fix two regex-related warnings:

    what_changed.py:273:0: W1401: Anomalous backslash in string: '\.'.
    String constant might be missing an r prefix.
    (anomalous-backslash-in-string)

There's still two warnings related to the 'sh' library. Those will be
fixed separately by removing it.

### DIFF
--- a/scripts/what_changed.py
+++ b/scripts/what_changed.py
@@ -6,9 +6,7 @@
 
 # A script to set labels on pull-rquests based on files being changed
 
-import sys
 import re, os
-from email.utils import parseaddr
 import sh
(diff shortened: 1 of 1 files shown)


ulfalizer:
Hmm... might as well fix the `sh` stuff in the same PR, or it won't pass.

ulfalizer:
Rolling these up into one PR. Gets messy otherwise.

PR #91, opened 2019-09-16
check_compliance.py: Run pylint on compare_footprint and size_report

ulfalizer:
These don't end in .py and so weren't being picked up. List them
manually. pylint will only run on them when they're modified, like for
the other Python scripts.

scripts/sanitycheck will be included later as well, but is currently
being overhauled.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -692,13 +692,17 @@ class PyLint(ComplianceTest):
         # Path to pylint configuration file
         pylintrc = os.path.join(os.path.dirname(__file__), "pylintrc")
 
-        # List of .py files added/modified by the commit(s).
+        # List of .py files added/modified by the commit(s). We also search for
+        # changes to some files that are known to be Python despite not ending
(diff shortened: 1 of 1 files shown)


marc-hb:
Asking the obvious question: why can't these be renamed? Would help not just `check_compliance.py` but any number of other tools too. Git does a decent job at tracking "pure renames" which are not (stupidly) polluted with many code changes in the same file and same commit. These files are also less than 500 lines each.

ulfalizer:
@marc-hb 
Might get a bit annoying if people ever run these directly (like `scripts/sanitycheck`), but other than that it's nicer.

Wonder if something outside the Zephyr repo calls them. `size_report` is run by `cmake/reports/CMakeLists.txt` in the Zephyr repo, but that's easy to fix at least.

@nashif 
Would anything break by renaming these?

ulfalizer:
Replaced with https://github.com/zephyrproject-rtos/ci-tools/pull/92 for now.

PR #95, opened 2019-09-18
add_info() test

ulfalizer:
Ignore for now.


ulfalizer:
Bleh... will need a new approach. Hadn't understood the test saving/loading stuff that the script does, and it breaks it.

PR #96, opened 2019-09-18
check_compliance.py: Always show warnings from checkpatch.pl

ulfalizer:
First commit adds support for informational messages from tests. Second commit uses it to report warnings from the `CheckPatch` test, even when there are no errors that would fail the test.

```
check_compliance.py: Add support for informational messages

Add a new ComplianceTest.add_info() function for showing informational
messages in tests without failing the test. Informational messages are
always shown on GitHub, regardless of whether the test succeeds or not.

CI runs check_compliance.py several times and keeps track of test
results by saving them to JUnit XML files that later get loaded, which
complicates things. To preserve informational messages, save them in a
new 'info_msg' attribute on <testcase> (added via the custom MyCase
class).

Piggyback removal of a declared-but-unused (never set) 'doc' attribute
on <testcase>, and some minor EDIT_TIP cleanup (add it in
github_comment(), since it should always appear).
```

```
check_compliance.py: Always show warnings from checkpatch.pl

Always show the output from checkpatch.pl on GitHub, even if there are
no errors. This way, warnings will always be seen.

checkpatch.pl produces no output if there are neither warnings nor
errors (because of --mailback), so this won't cause GitHub spam.

Fixes: #63
```

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -76,11 +76,13 @@ def get_shas(refspec):
 
 class MyCase(TestCase):
     """
-    Implementation of TestCase specific to our tests.
-
+    Custom junitparser.TestCase for our tests that adds some extra <testcase>
+    XML attributes. These will be preserved when tests are saved and loaded.
     """
     classname = Attr()
-    doc = Attr()
+    # Remembers informational messages. These can appear on successful tests
+    # too, where TestCase.result isn't set.
+    info_msg = Attr()
 
 
 class ComplianceTest:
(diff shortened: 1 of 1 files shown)


marc-hb:
Nice! The code looks pretty straight-forward. Would it be possible to see a "demo" somewhere?

ulfalizer:
@marc-hb 
See above for demo. I was hoping to have it failure-free too, but I messed up.~

ulfalizer:
Hmm... the behavior is slightly wonky when there's an immediate success along with an informational message though. Get a `All checks passed` that then gets edited to `All checks are passing now`.

It's due to `check_compliance.py` being run twice by the `ci-pipelines` repo (once for everything-but-documentation, and then for documentation, with saving tests in-between). Second pass now sees the message that was posted by the first pass, assumes it's a failure message, and edits it.

Think I could work around it... w8.

ulfalizer:
Seems to be working now. That message thing gets a bit confusing.

PR #97, opened 2019-09-19
check_compliance.py: don't abort on DeviceTreeCheck failures.

marc-hb:
Catch Exception not SystemExit.

Also use logger.info() to show which particular test print warning(s)
and how to run them directly.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -513,7 +513,9 @@ class DeviceTreeCheck(ComplianceTest):
         scripts_path = os.path.join(ZEPHYR_BASE, "scripts", "dts")
 
         sys.path.insert(0, scripts_path)
+        logger.info("cd %s && ./testdtlib.py" % scripts_path)
         import testdtlib
+        logger.info("cd %s && ./testedtlib.py" % scripts_path)
         import testedtlib
 
         # Hack: The test suites expect to be run from the scripts/dts
@@ -528,7 +530,7 @@ class DeviceTreeCheck(ComplianceTest):
         try:
             testdtlib.run()
             testedtlib.run()
-        except SystemExit as e:
+        except Exception as e:
             # The dtlib and edtlib test suites call sys.exit() on failure,
             # which raises SystemExit. Let any errors in the test scripts
             # themselves trickle through and turn into an internal CI error.


scripts/check_compliance.py
```diff
@@ -513,7 +513,9 @@ def run(self):
         scripts_path = os.path.join(ZEPHYR_BASE, "scripts", "dts")
 
         sys.path.insert(0, scripts_path)
+        logger.info("cd %s && ./testdtlib.py" % scripts_path)
```
ulfalizer: Think you could make this

```python
logger.info("cd %s && ./testedtlib.py", scripts_path)
```

The `logger` module has some built-in string formatting stuff.

scripts/check_compliance.py
```diff
@@ -528,7 +530,7 @@ def run(self):
         try:
             testdtlib.run()
             testedtlib.run()
-        except SystemExit as e:
+        except Exception as e:
```
ulfalizer: Catching `SystemExit` here is deliberate, because that's how the test suites reports failures (they just call `sys.exit()`, which raises `SystemExit`, which isn't a subclass of `Exception`).

Are you planning to make some other changes?

scripts/check_compliance.py
```diff
@@ -513,7 +513,9 @@ def run(self):
         scripts_path = os.path.join(ZEPHYR_BASE, "scripts", "dts")
 
         sys.path.insert(0, scripts_path)
+        logger.info("cd %s && ./testdtlib.py" % scripts_path)
         import testdtlib
+        logger.info("cd %s && ./testedtlib.py" % scripts_path)
```
ulfalizer: Maybe it's better to move these comments next to the `run()` calls below. That's where they actually get run.

scripts/check_compliance.py
marc-hb: Yet catching SystemExit aborts immediately and doesn't run the other checks while catching Exception works.... how come? Have you tested this? I know, I know, who ever tests error handling? :-)

> Are you planning to make some other changes?

None related to this.

ulfalizer: Hmm... if `SystemExit` isn't caught here, then it should trickle all the way up to `main()`, where it'd cause an internal `check_compliance.py` error instead.

An earlier version accidentally catched `Exception` here instead, and that's what I saw then when tests failed (it's one of the thing that motivated me to add the reporting of internal errors to GitHub, because it was a PITA to debug :).

Are you seeing something else?

marc-hb: Please just try this:
```shell
git stash
git fetch upstream/master
git checkout 79d626f5c7cc
```
Now break one test one purpose:
```diff
--- a/scripts/dts/testedtlib.py
+++ b/scripts/dts/testedtlib.py
@@ -37,7 +37,7 @@ def run():
     #
 
     verify_streq(edt.get_dev("/interrupt-parent-test/node").interrupts,
-                 "[<Interrupt, name: foo, target: <Device /interrupt-parent-test/controller in 'test.dts', binding test-bindings/interrupt-3-cell.yaml>, specifier: {'one': 1, 'two': 2, 'three': 3}>, <Interrupt, name: bar, target: <Device /interrupt-parent-test/controller in 'test.dts', binding test-bindings/interrupt-3-cell.yaml>, specifier: {'one': 4, 'two': 5, 'three': 6}>]")
+                 "[<Interrupt, name: foo, target: <Device /interrupt-parent-test/controller in 'test.dts', binding XXXtest-bindings/interrupt-3-cell.yaml>, specifier: {'one': 1, 'two': 2, 'three': 3}>, <Interrupt, name: bar, target: <Device /interrupt-parent-test/controller in 'test.dts', binding test-bindings/interrupt-3-cell.yaml>, specifier: {'one': 4, 'two': 5, 'three': 6}>]")
 
     verify_streq(edt.get_dev("/interrupts-extended-test/node").interrupts,
```
and now run check_compliance.py. It aborts before running other tests. Apply this PR, problem goes away.

That's all I know.

ulfalizer: Ohh... I see what it is now. `testedtlib.py` does this:

```python
def fail(msg):
    raise Exception("test failed: " + msg)
``` 

It should be doing this instead, to match what `check_compliance.py` expects:

```python
def fail(msg):
    sys.exit("test failed: " + msg)
```

The bug is in `testedtlib.py`. Sorry about that.

PR #98, opened 2019-09-19
check_compliance.py: show where PyLint is actually run

marc-hb:
Yes, check_compliance.py can run pylint on itself.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -721,7 +721,7 @@ class PyLint(ComplianceTest):
     _doc = "https://www.pylint.org/"
 
     def run(self):
-        self.prepare(ZEPHYR_BASE)
+        self.prepare(os.getcwd())
 
         # Path to pylint configuration file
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -1057,6 +1057,10 @@ def github_comment(msg):
         github_pr.create_issue_comment(msg)
 
 
+# Cache used by get_bot_comment(). Use 0 instead of None for "no cached value"
+# so that None (no comment) can be cached.
+cached_bot_comment = 0
+
```
ulfalizer: Could put two blanks lines here, just as some PEP 8 zealotry.

marc-hb:
Actually, it's not that simple: you need to `unset ZEPHYR_BASE` to run it on itself. Back to the drawing board.

marc-hb:
See PR #100 for running it in GIT_TOP

PS: PR number #100! Is there a prize?

PR #99, opened 2019-09-19
check_compliance.py: Preserve 'info_msg' better in XML

ulfalizer:
Tabs were getting lost, causing the bot to edit the checkpatch message
to remove them the second time check_compliance.py was run (it's run
multiple times by CI, with test results saved to XML between runs). The
bot edits the comment when the text changes.

This isn't my area, but properly escaping and unescaping text for XML
attributes seems surprisingly tricky. There's
xml.sax.saxutils.quoteattr() for escaping, but there doesn't seem
anything built-in for undoing it.

Just use 'unicode-escape' instead, which should be good enough in
practice. It replaces newlines and tabs with the two-character strings
\n and \t, etc.

junitparser already deals with quotes and stuff.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -174,14 +174,17 @@ class ComplianceTest:
 
         Can be called many times within the same test to add multiple messages.
         """
-        # Hack to preserve newlines in the attribute when tests are saved to
-        # .xml and reloaded. junitparser doesn't seem to handle it correctly.
-        msg = msg.replace("\n", "[newline]")
+        def escape(s):
+            # Hack to preserve e.g. newlines and tabs in the attribute when
+            # tests are saved to .xml and reloaded. junitparser doesn't seem to
+            # handle it correctly, though it does escape stuff like quotes.
+            # unicode-escape replaces newlines with \n (two characters), etc.
+            return s.encode("unicode-escape").decode("utf-8")
 
         if not self.case.info_msg:
-            self.case.info_msg = msg
+            self.case.info_msg = escape(msg)
         else:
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -1002,12 +1005,14 @@ def set_commit_status(status, msg):
...
-        # an explanation of the [newline] thing.
+        # Always show informational messages. See ComplianceTest.add_info() re.
+        # the decoding thing. We undo it here.
         if case.info_msg:
             comment += "## {} (informational only, not a failure)\n\n" \
                        "```\n{}\n```\n\n".format(
-                           case.name, case.info_msg.replace("[newline]", "\n"))
+                           case.name,
+                           case.info_msg.encode("utf-8")
+                                        .decode("unicode-escape"))
```
marc-hb: In the future a `get_info()` accessor would be more consistent.

ulfalizer:
> Mildly worried it may make `compliance.xml` less readable. Or more? @ulfalizer could you copy/paste an example here?

Looks like

```
<testcase classname="Guidelines" info_msg="foo\n\tbar\n\t\tbaz" name="License">
```

Gets longer for checkpatch messages though, obviously.

It seems you can add custom elements with junitparser too. Might make things more complicated though, and it seems kinda obscure (the documentation has "there may be once in 1000 years you want to it this way, but anyways").

scripts/check_compliance.py
ulfalizer: Yeah, will def. add a helper for it if it needs to be unescaped in more places.

PR #100, opened 2019-09-19
pylint fixes: don't ignore fatal issues, run in GIT_TOP, logging

marc-hb:
See commit messages.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -723,7 +723,7 @@ class PyLint(ComplianceTest):
     _doc = "https://www.pylint.org/"
 
     def run(self):
-        self.prepare(ZEPHYR_BASE)
+        self.prepare(GIT_TOP)
 
         # Path to pylint configuration file
         pylintrc = os.path.join(os.path.dirname(__file__), "pylintrc")
@@ -738,8 +738,10 @@ class PyLint(ComplianceTest):
             ":!boards/xtensa/intel_s1000_crb/support/create_board_img.py") \
             .splitlines()
 
-        # Filter out everything but Python files
-        py_files = filter_py(files)
+        # Filter out everything but Python files. Keep filenames
+        # relative (to GIT_TOP) to stay farther from any command line
+        # limit.
+        py_files = filter_py(GIT_TOP, files)
         if not py_files:
             return
 
@@ -751,7 +753,7 @@ class PyLint(ComplianceTest):
                 pylintcmd,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
-                cwd=ZEPHYR_BASE)
+                cwd=GIT_TOP)
         except FileNotFoundError:
             self.error("pylint not found. Check that the directory it's in is "
                        "listed in the PATH environment variable.")
@@ -764,15 +766,17 @@ class PyLint(ComplianceTest):
             self.add_failure(stdout.decode("utf-8") + stderr.decode("utf-8"))
 
 
-def filter_py(fnames):
+def filter_py(root, fnames):
     # PyLint check helper. Returns all Python script filenames among the
-    # filenames in 'fnames'. Uses the python-magic library, so that we can
-    # detect Python files that don't end in .py as well. python-magic is a
-    # frontend to libmagic, which is also used by 'file'.
+    # filenames in 'fnames', relative to directory 'root'. Uses the
+    # python-magic library, so that we can detect Python files that
+    # don't end in .py as well. python-magic is a frontend to libmagic,
+    # which is also used by 'file'.
 
     return [fname for fname in fnames
             if fname.endswith(".py") or
-               magic.from_file(fname, mime=True) == "text/x-python"]
+               magic.from_file(os.path.join(root, fname),
+                               mime=True) == "text/x-python"]
 
 
 class License(ComplianceTest):
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -1061,6 +1061,11 @@ def github_comment(msg):
         github_pr.create_issue_comment(msg)
 
 
+# Cache used by get_bot_comment(). Use 0 instead of None for "no cached value"
+# so that None (no comment) can be cached.
+cached_bot_comment = 0
+
+
 def get_bot_comment():
     # Returns any previous comment posted by the bot in 'github_pr', or None if
     # the bot hasn't posted any comment (or there's no pull request)
@@ -1083,11 +1088,6 @@ def get_bot_comment():
     if cached_bot_comment == 0:
         cached_bot_comment = get_comment()
     return cached_bot_comment
-
-
-# Cache used by get_bot_comment(). Use 0 instead of None for "no cached value"
-# so that None (no comment) can be cached.
-cached_bot_comment = 0
 
 
 def parse_args():
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -526,7 +526,9 @@ class DeviceTreeCheck(ComplianceTest):
         old_dir = os.getcwd()
         os.chdir(scripts_path)
         try:
+            logger.info("cd %s && ./testdtlib.py", scripts_path)
             testdtlib.run()
+            logger.info("cd %s && ./testedtlib.py", scripts_path)
             testedtlib.run()
         except SystemExit as e:
             # The dtlib and edtlib test suites call sys.exit() on failure,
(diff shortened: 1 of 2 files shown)


scripts/pylintrc
```diff
@@ -59,6 +59,8 @@ score=no
 disable=all
 # Identifiers are in the same order as in 'pylint3 --list-msgs'
 enable=
+    fatal,
+#    E, # maybe some day
```
ulfalizer: Did you check that this doesn't accidentally disable all the tests that follow (`empy-docstring`, etc.)?

`pylint` seems to silently ignore stuff it doesn't understand here sometimes.

ulfalizer: Should also put `fatal` where it appears in the output of `pylint --list-msgs` (see the comment above the `enable=` line).

What's E btw?

scripts/check_compliance.py
```diff
@@ -737,17 +739,18 @@ def run(self):
...
         if not py_files:
             return
 
+        pylintcmd = ["pylint", "--rcfile=" + pylintrc] + py_files
+        logger.info(" ".join([shlex.quote(word) for word in pylintcmd]))
         try:
             # Run pylint on added/modified Python files
-            process = subprocess.Popen(
-                ["pylint", "--rcfile=" + pylintrc] + py_files,
+            process = subprocess.Popen(pylintcmd,
```
ulfalizer: `pylintcmd` should go on a separate line, so that the second argument isn't less indented than the first argument.

scripts/pylintrc
ulfalizer: IIRC, I tried to enable `fatal`, but it causes a bunch of issues to due modules not being found, etc.

ulfalizer: Then again, might be nice to do a test run with it, to try to fix the cases that can be fixed. Would be nice if we could get all of them fixed, so that everything can be linted.

marc-hb: > Did you check that this doesn't accidentally disable all the tests that follow (empy-docstring, etc.)?

Yes, this syntax is valid.

> What's E btw? 

errors, see the bottom of `man pylint`. Unlike `fatal`, only the initial works with pylint 2.1.1. No idea why.

> IIRC, I tried to enable fatal, but it causes a bunch of issues to due modules not being found, etc.

These are errors, you've been testing `E` by mistake. They are the reason why I commented it out.

ulfalizer: > errors, see the bottom of `man pylint`. Unlike `fatal`, only the initial works with pylint 2.1.1. No idea why.

Oh... didn't know that `--enable` accepted categories and stuff too, but seems it does. Got confused since it's not listed in `--list-msgs`, but it's a different thing then.

> These are errors, you've been testing `E` by mistake. They are the reason why I commented it out.

Hmm... what do you mean by "testing `E` by mistake"? It was never enabled.

Did you notice any stuff that could be fixed with the E category enabled?

marc-hb: > Did you notice any stuff that could be fixed with the entire E category enabled?

Not sure what you mean exactly by "could be fixed" but yes, enabling the entire Error category does report extra errors. I only gave Error a quick try and didn't really pay much attention to them out of "yak shaving" fear: the only purpose of this series is to make it easy to run `check_compliance.py` on `ci-tools` itself + fix what made that difficult along the way (which is already some yak shaving depending on how you look at it)

Now if you're interested, looking at all Errors is as easy as `pylint --rcfile=tools/ci-tools/scripts/pylintrc $(git ls-files \*.py)`. The two most common types seem to be E0401 and E1101
```
************* Module device
boards/xtensa/intel_s1000_crb/support/device.py:17:0: E0401: Unable to import 'periphery' (import-error)
************* Module genrest
doc/scripts/genrest.py:110:23: E1101: Instance of 'str' has no 'prompt' member (no-member)
```

marc-hb: > Hmm... what do you mean by "testing E by mistake"?

I mean enabling "fatal" doesn't show any extra... fatality for me (except when bending check_compliance.py over backwards, see commit message). You?

ulfalizer: Sorry, hadn't checked the commit message. I get it now. I had misunderstood the purpose of `fatal`.

Not making excuses, but this is why I always paste all the commit messages into the PR message, because I assume other people have my flaws too. :)

scripts/check_compliance.py
```diff
@@ -741,10 +743,11 @@ def run(self):
         if not py_files:
             return
 
+        pylintcmd = ["pylint", "--rcfile=" + pylintrc] + py_files
+        logger.info(" ".join([shlex.quote(word) for word in pylintcmd]))
```
ulfalizer: Could shorten a tiny bit with a generator:

```python
logger.info(" ".join(shlex.quote(word) for word in pylintcmd))
```

scripts/pylintrc
```diff
@@ -59,6 +59,8 @@ score=no
 disable=all
 # Identifiers are in the same order as in 'pylint3 --list-msgs'
 enable=
+    fatal,
```
ulfalizer: Should be put so that it appears where `pylint --list-msgs` lists it, for consistency.

scripts/pylintrc
ulfalizer: Think this comment might be confusing to people. Maybe it could be removed. Could put a longer comment somewhere else that explains what E is, why it can't be enabled, etc., instead.

scripts/check_compliance.py
```diff
@@ -737,17 +739,18 @@ def run(self):
             .splitlines()
 
         # Filter out everything but Python files
-        py_files = filter_py(files)
+        py_files = filter_py(GIT_TOP, files)
```
ulfalizer: Could do this maybe, so that `filter_py()` can keep taking just a list of paths (otherwise, the doc comment for it should be updated).

```python
py_files = filter_py(os.path.join(GIT_TOP, path) for path in files)
```

ulfalizer: Maybe the `git diff` could be run in `GIT_TOP` too, even though it doesn't matter for the output (since the current directory is assumed to be inside the repository you want to check). Might be less confusing.

scripts/pylintrc
marc-hb: ok I'll remove it.

marc-hb: I didn't copy/paste all the commit messages partly for laziness and as a small and impulsive act of rebellion against the generally negligent way github deals with commit messages. For instance they can't even be reviewed. At the very least github could pre-can the description with that copy/paste itself.

On the other hand, the only thing I put in the description is: "See commit messages" :-)

mbolivar:
> I can remove it if you insist.

I do not insist. Thanks for the explanation. All good as far as I'm concerned, +1 to merge.

scripts/check_compliance.py
marc-hb: I've considered running `git diff` in `GIT_TOP`, however it gives the false impression that it makes a difference - it doesn't. It does only when using `git diff --relative`.

scripts/pylintrc
mbolivar: I think removing it is less preferable to keeping it with an explanation of what it would do if uncommented in this case. I'm normally dead set against commented-out code, but your point that it's confusing syntax at the bottom of a manpage is well taken.

scripts/pylintrc
marc-hb: Thanks, you finally made me realize something important! I've been confusing `fatal=F001` with `F` the whole time. Now my excuse is: as of pylint 2.1.1 they behave exactly the same because there is no F002, F003,... at least not yet.

This explains why `E` works while `error` doesn't exist.

So here's my latest proposal:
- change `fatal` to `F # atal`
- keep it first because it's not just ant error appearing in `--list-msgs` but an "entire" level
- remove `# E` because this hint is not needed any more

Thoughts?

ulfalizer: Sounds good to me, as long as the Zephyr repo doesn't trigger any F stuff.

ulfalizer: Could update the comment above `enable=` to mention that categories are listed first too. Something like

```
# Identifiers are in the same order as in 'pylint --list-msgs'. Categories of
# errors (like F) are listed first.
```

Hints a bit re. what F is too.

scripts/check_compliance.py
ulfalizer: Alright, fine with me.

scripts/pylintrc
ulfalizer: Yeah, I'll just blame GitHub for making me lazy by not making it easier to view all the commit messages. ;)

Would be nice if they popped up in some window you could scroll through or something.

scripts/check_compliance.py
marc-hb: > (otherwise, the doc comment for it should be updated).

I went for updating the filter_py() comment(s) to stay farther from any command line limit.

ulfalizer: Alright. Will get the same list from `filter_py()` in both cases though.

carlescufi:
@thst-nordic could you please review in the context of a downstream using this?

marc-hb:
> @thst-nordic could you please review in the context of a downstream using this?

@carlescufi which particular change are you concerned about? I could extract it not to block the other, minor fixes (to be honest I'm struggling to guess which particular commit is not minor...)

PR #106, opened 2019-11-06
what_changed: make it do nothing

galak:
Replace with a nop so we test using the GH labeler workflow.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


galak:
> Lets keep the code and just disable it in the pipeline.

Doing this for just testing purposes.

nashif:
> Lets keep the code and just disable it in the pipeline.

we can remove it later once Actions is working properly.

PR #107, opened 2019-11-07
Revert "check_compliance.py: Make license check informational"

galak:
This reverts commit 747795b933555036af09ed7de7e76900aaec7b3b.

We want the license check to error out.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


galak:
@ulfalizer I merged your change thinking it was just changing the text of the message.  We actually want the license check to be treated as an error.

PR #116, opened 2020-01-03
check_compliance.py: Kconfig: Flag redundant $ZEPHYR_BASE in 'source'

ulfalizer:
'source's like

    source "$(ZEPHYR_BASE)/Kconfig.zephyr"

can be simplified to

    source "Kconfig.zephyr"

since $srctree already points to the Zephyr root.

Flag it in the `Nits` test.

This also avoids absolute paths showing up in some places.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -684,20 +684,21 @@ class Nits(ComplianceTest):
             is_kconfig = "Kconfig" in fname
 
             if is_kconfig:
-                self.check_kconfig_file(fname)
+                self.check_kconfig_header(fname)
+                self.check_redundant_zephyr_source(fname)
 
             if fname.endswith((".c", ".cpp", ".h", ".ld", ".py", ".rst",
                                ".yaml", ".yml")) or is_kconfig:
                 self.check_source_file(fname)
 
-    def check_kconfig_file(self, fname):
-        # Nits related to Kconfig files
+    def check_kconfig_header(self, fname):
+        # Checks for a spammy copy-pasted header format
 
         with open(os.path.join(GIT_TOP, fname), encoding="utf-8") as f:
             contents = f.read()
 
         # 'Kconfig - yada yada' has a copy-pasted redundant filename at the
-        # start. This probably means all of the header was copy-pasted.
+        # top. This probably means all of the header was copy-pasted.
         if re.match(r"\s*#\s*(K|k)config[\w.-]*\s*-", contents):
             self.add_failure("""
 Please use this format for the header in '{}' (see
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -715,6 +716,22 @@ def check_kconfig_file(self, fname):
 failure.
 """.format(fname))
 
+    def check_redundant_zephyr_source(self, fname):
+        # Checks for 'source "$(ZEPHYR_BASE)/Kconfig[.zephyr]"', which can be
+        # be simplified to 'source "Kconfig[.zephyr]"'
+
+        with open(os.path.join(GIT_TOP, fname), encoding="utf-8") as f:
```
marc-hb: GIT_TOP or ZEPHYR_BASE here? I don't have a clue, cause I don't know if/how Kconfig interacts with other west modules.

- Can a single Kconfig be split across several west modules?
- Can a west module have its own, standalone Kconfig independent of 
- Both at the same time? (for different modules)
Is (some of) this already documented anywhere? @mbolivar , @carlescufi, others?

The answers to these questions should probably not block this PR, I just think it's a good opportunity to ask.

ulfalizer: Kconfig itself doesn't know about modules. The way it works is that `scripts/zephyr_module.py` generates a file that includes the Kconfig files for the modules:

    source "/path/to/some/module/Kconfig"
    source "/path/to/some/other/module/Kconfig"
    ...

That file then gets included from `Kconfig.zephyr`:

```
menu "Modules"

source "$(CMAKE_BINARY_DIR)/Kconfig.modules"
source "modules/Kconfig"

endmenu
```

I think `GIT_TOP` makes more sense than `ZEPHYR_BASE`, because `source "$(ZEPHYR_BASE)/Kconfig"` is guaranteed to be redundant even if you're in a module. If `$srctree` wasn't set to `$ZEPHYR_BASE`, then all the `source`s within the Zephyr repo would fail, so `$srctree = $ZEPHYR_BASE` is guaranteed.

The paths to the changed files are relative to `GIT_TOP` too, since they're generated with this:

```
for fname in git("diff", "--name-only", "--diff-filter=d",
                 COMMIT_RANGE).splitlines():
```

marc-hb: > The way it works is that scripts/zephyr_module.py generates a file 

Perfect summary thanks, it triggered some [old memories](https://github.com/zephyrproject-rtos/zephyr/pull/15042) :-)
This summary or something like it should be in the documentation if not yet, maybe added to this section? https://docs.zephyrproject.org/latest/guides/build/index.html
(Funny enough the "Kconfig" and "west" section are currently one after the other)

> I think GIT_TOP makes more sense than ZEPHYR_BASE, because source "$(ZEPHYR_BASE)/Kconfig" is guaranteed to be redundant even if you're in a module.

In other words, better offer the flexibility to check files in any random GIT_TOP? Makes sense, thanks.

Can you change this reference to GIT_TOP and probably the previous one to `path_hint`? This will make sure they never go out of sync with each other.

ulfalizer: > Can you change this reference to GIT_TOP and probably the previous one to `path_hint`? This will make sure they never go out of sync with each other.

Don't think they'd ever get out of sync here, since it's inherent in the test that it checks all the changed files in the repo. Would need to change more stuff (e.g. the working directory for the `git diff` command) to make it possible to change it to a subdirectory of the repository.

Might make more sense for other tests. Can probably skip the indirection here, since it's not really customizable in the code.

PR #119, opened 2020-01-21
check_compliance.py: Make compatible with pre-DEVICETREE_CONF renaming

ulfalizer:
Older trees don't have the GENERATED_DTS_BOARD ->
DEVICETREE_CONF renaming yet. Set GENERATED_DTS_BOARD too in the
KconfigCheck test, to make it compatible.


anangl:
Could you rebase it on master, so that it contains also commit https://github.com/zephyrproject-rtos/ci-tools/commit/8a70a13e23bbcf8f9eb2ac58fdbb027acc8ca38e and can be tested by https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1750?

PR #124, opened 2020-01-27
check_compliance.py: Avoid running 'git' before _main()

ulfalizer:
Running 'git rev-parse --show-toplevel' to set GIT_TOP before _main()
prevents any errors from it from being reported to GitHub, because the
GitHub connection hasn't been initialized yet.

If there's an error, the error reporting is currently broken as well,
because git() calls err(), which is defined at the end of the file.

Initialize GIT_TOP in _main() instead of at the top level, and use a
magic "\<repo\>" string to refer to the top-level repository directory
when setting the 'ComplianceTest.path_hint' class variables instead.

Discovered while working on some unrelated error reporting improvements.

Also move the ZEPHYR_BASE initialization to a better spot, next to other
global variables.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -28,6 +28,9 @@ EDIT_TIP = "\n\n*Tip: The bot edits this comment instead of posting a new " \
 
 logger = None
 
+# This ends up as None when we're not running in a Zephyr tree
+ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE')
+
 
 def git(*args, cwd=None):
     # Helper for running a Git command. Returns the rstrip()ed stdout output.
@@ -57,13 +60,6 @@ def git(*args, cwd=None):
 {}""".format(git_cmd_s, git_process.returncode, stdout, stderr))
 
     return stdout.rstrip()
-
-
-# This ends up as None when we're not running in a Zephyr tree
-ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE')
-
-# The absolute path of the top-level git directory
-GIT_TOP = git("rev-parse", "--show-toplevel")
 
 
 def get_shas(refspec):
@@ -103,6 +99,12 @@ class ComplianceTest:
     path_hint:
       The path the test runs itself in. This is just informative and used in
       the message that gets printed when running the test.
+
+      The magic string "<git-top>" refers to the top-level repository
+      directory. This avoids running 'git' to find the top-level directory
+      before main() runs (class variable assignments run when the 'class ...'
+      statement runs). That avoids swallowing errors, because main() reports
+      them to GitHub.
     """
     def __init__(self):
         self.case = MyCase(self.name)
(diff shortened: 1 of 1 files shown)


scripts/check_compliance.py
```diff
@@ -1295,8 +1300,8 @@ def _main(args):
 
         test = testcase()
         try:
-            print("Running {:16} tests in {} ..."
-                  .format(test.name, test.path_hint))
+            print(f"Running {test.name:16} tests in "
+                  f"{GIT_TOP if test.path_hint == '<repo>' else test.path_hint} ...")
```
marc-hb: I don't get this. If GIT_TOP was successfully found, why was `path_hint` not already initialized?

scripts/check_compliance.py
```diff
@@ -103,6 +99,10 @@ class ComplianceTest:
     path_hint:
       The path the test runs itself in. This is just informative and used in
       the message that gets printed when running the test.
+
+      The magic string "<repo>" is replaced with the top-level repository
+      directory. This avoids running 'git' before _main(), which would prevent
+      errors from it from being reported to GitHub.
```
marc-hb: Can you rephrase a bit to briefly mention that `_main()` is the one reporting to github? I forgot that and was really struggling.

scripts/check_compliance.py
ulfalizer: `path_hint` is a class variable, so it gets initialized when the class is defined (when the Python interpreter runs the `class ...` statement).

This is the only place where `path_hint` is used, so rewriting it from `"<git-top>"` (new name) to `GIT_TOP` seemed unnecessary. Nicer to have it stateless I think.

scripts/check_compliance.py
ulfalizer: Rephrased and added some more details:

```
The magic string "<git-top>" refers to the top-level repository directory. This
avoids running 'git' to find the top-level directory before main() runs (class
variable assignments run when the 'class ...' statement runs). That avoids
swallowing errors, because main() reports them to GitHub.
```

Is that okay?

ulfalizer:
> I find `<repo>` loses the important "TOP" part of "GIT_TOP". Why not just `@git_top@`?

Renamed it to `<git-top>`.

> Besides no .git/ subdirectory, I'm curious what other sort of errors `rev-parse --show-top-level` can encounter. What did you experience?

I just put an error into the `git` command to test https://github.com/zephyrproject-rtos/ci-tools/pull/125. Makes it fail with an unrelated Python error, because the `err()` method isn't defined yet.

Means that if there was an error, it wouldn't even be properly reported when running locally.

More a general robustness thing so that the script never swallows errors.

scripts/check_compliance.py
marc-hb: Sorry for the previous comment, it assumed you _were_ rewriting path_hint.

As I wrote in the main thread, I think it would be better to rename `path_hint` to `directory_tested` or something like it and then use that to unhardcode `cwd` in `subprocess.check_output(cwd=directory_tested,... )` in `GitLint`,`CheckPatch` and maybe others I missed.

This would be added a couple lines above:
```python
if testcase.directory_tested == "<git-top>":
      testcase.directory_tested = GIT_TOP
```

PR #126, opened 2020-01-28
check_compliance.py: Clean up get_defined_syms() a bit

ulfalizer:
Can do a single regex scan on the 'git grep' output and use
--perl-regexp instead of --extended-regexp to get non-capturing groups.
Also shortens things a bit.

### DIFF
--- a/scripts/check_compliance.py
+++ b/scripts/check_compliance.py
@@ -427,19 +427,18 @@ def get_defined_syms(kconf):
     # definitions. Doing it "properly" with Kconfiglib is still useful for the
     # main tree, because some symbols are defined using preprocessor macros.
 
-    # Warning: Needs to work with both --extended-regexp and the 're' module
-    regex = r"^\s*(menu)?config\s*([A-Z0-9_]+)\s*(#|$)"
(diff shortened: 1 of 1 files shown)


marc-hb:
> use
--perl-regexp instead of --extended-regexp to get non-capturing groups.

marc-hb:
> use --perl-regexp instead of --extended-regexp to get non-capturing groups.

Reviewing this level of detail is too time-consuming, I'll pass on this one sorry :-(

ulfalizer:
> > use --perl-regexp instead of --extended-regexp to get non-capturing groups.
> 
> Reviewing this level of detail is too time-consuming, I'll pass on this one sorry :-(

It's tested and works at least.

Non-capturing groups are like `()` in regexes, except you can't fetch the contents of them. They're written like `(?:...)`. So in `regex = r"^\s*(?:menu)?config\s*([A-Z0-9_]+)\s*(?:#|$)"`, only the `([A-Z0-9_]+)` part will be captured.

They're handy there because it makes `re.findall()` just return just a list with the stuff matched by `([A-Z0-9_]+)`, instead of getting a list of tuples with all the groups.

Maybe it was too much detail for the commit message though...

PR #128, opened 2020-02-07
check_compliance.py: Check for duplicated dependencies on Kconfig syms

ulfalizer:
Check for e.g. 'if FOO' within 'if FOO', or 'depends on FOO' within 'if
FOO'. Such duplicated dependencies make it easier to mess up changes to
Kconfig files (because it's not obvious that the same dependency is
added twice), and make the autogenerated documentation uglier.

Some of them might have been added due to Kconfig misunderstandings too.


ulfalizer:
Marking DNM until I've fixed the stuff it flags in Zephyr.

PR #134, opened 2020-03-23
compliance: blacklist anonymous github emails

mbolivar-nordic:
We need real addresses to be able to contact authors of code.


mbolivar-nordic:
Looks like this is working, as tested in https://github.com/zephyrproject-rtos/zephyr/pull/23697. If this gets merged I'll update the Zephyr PR with the corresponding SHA.

